PR #30, opened 2019-06-30
Implement Expect extension 'throwsWithMessage'

mprokushenkov:


tests/UnitTests/Expect.fs
```diff
@@ -0,0 +1,27 @@
+﻿namespace Pulsar.Client.UnitTests
+
+open System
+open Expecto
+open Expecto.Flip
+
+module internal Expect =
+
+    let throwsTh<'texn> f =
```
Lanayx: Can you please rename it or make private? I can't guess what throwsTh means =(

mprokushenkov: Ok. Now it named 'throwsT2'. Because original Expecto function has name 'throwsT' and 'throwsT2' extends 'throwsT' behavior ,'throwsT2' follows same naming pattern.

PR #46, opened 2019-07-25
split PulsarCommand DU

vilinski:
Splitted valid and invalid command DU cases

Removed try/catch from parsing function because:
1. catched exception is thrown later
2. exception context would be lost, because it is not `reraise` aka C#'s `throw;`,

### DIFF
--- a/src/Pulsar.Client/Internal/ClientCnx.fs
+++ b/src/Pulsar.Client/Internal/ClientCnx.fs
@@ -123,6 +123,40 @@ type ClientCnx (broker: Broker,
         loop ()
     )
 
+    let wrapCommand readMessage (command: BaseCommand) =
+        match command.``type`` with
+        | BaseCommand.Type.Connected ->
+            Ok (XCommandConnected command.Connected)
+        | BaseCommand.Type.PartitionedMetadataResponse ->
+            Ok (XCommandPartitionedTopicMetadataResponse command.partitionMetadataResponse)
+        | BaseCommand.Type.SendReceipt ->
+            Ok (XCommandSendReceipt command.SendReceipt)
+        | BaseCommand.Type.Message ->
+           let metadata,payload = readMessage()
+           Ok (XCommandMessage (command.Message, metadata, payload))
+        | BaseCommand.Type.LookupResponse ->
+            Ok (XCommandLookupTopicResponse command.lookupTopicResponse)
+        | BaseCommand.Type.Ping ->
+            Ok (XCommandPing command.Ping)
+        | BaseCommand.Type.ProducerSuccess ->
+            Ok (XCommandProducerSuccess command.ProducerSuccess)
+        | BaseCommand.Type.Success ->
+            Ok (XCommandSuccess command.Success)
+        | BaseCommand.Type.SendError ->
+            Ok (XCommandSendError command.SendError)
+        | BaseCommand.Type.CloseProducer ->
+            Ok (XCommandCloseProducer command.CloseProducer)
+        | BaseCommand.Type.CloseConsumer ->
+            Ok (XCommandCloseConsumer command.CloseConsumer)
+        | BaseCommand.Type.ReachedEndOfTopic ->
+            Ok (XCommandReachedEndOfTopic command.reachedEndOfTopic)
+        | BaseCommand.Type.GetTopicsOfNamespaceResponse ->
+            Ok (XCommandGetTopicsOfNamespaceResponse command.getTopicsOfNamespaceResponse)
+        | BaseCommand.Type.Error ->
+            Ok (XCommandError command.Error)
+        | unknownType ->
+            Result.Error (UnknownCommandType unknownType)
+
     let tryParse (buffer: ReadOnlySequence<byte>) readerId =
         let array = buffer.ToArray()
         if (array.Length >= 8)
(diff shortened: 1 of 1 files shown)


src/Pulsar.Client/Internal/ClientCnx.fs
```diff
@@ -123,73 +125,68 @@ type ClientCnx (broker: Broker,
 
     let tryParse (buffer: ReadOnlySequence<byte>) readerId =
         let array = buffer.ToArray()
-        if (array.Length >= 8)
-        then
+        if (array.Length <= 8)
```
Lanayx: = case forgotten

src/Pulsar.Client/Internal/ClientCnx.fs
```diff
@@ -123,73 +125,68 @@ type ClientCnx (broker: Broker,
...
-                    | BaseCommand.Type.GetTopicsOfNamespaceResponse ->
-                        XCommandGetTopicsOfNamespaceResponse (command.getTopicsOfNamespaceResponse, buffer.GetPosition(int64 frameLength))
-                    | BaseCommand.Type.Error ->
-                        XCommandError (command.Error, buffer.GetPosition(int64 frameLength))
-                    | _ as unknownCommandType ->
-                        InvalidCommand (Exception(sprintf "Unknown command type: '%A'" unknownCommandType))
-                with
-                | ex ->
-                    InvalidCommand ex
+            if (array.Length <= frameLength)
```
Lanayx: = case forgotten

src/Pulsar.Client/Internal/ClientCnx.fs
```diff
@@ -212,73 +209,75 @@ type ClientCnx (broker: Broker,
...
-                    | InvalidCommand ex ->
-                        raise ex
+                    | Result.Error (UnknownCommandType unknownType) ->
+                        failwithf "Unknown command type %A" unknownType
+                    | Result.Ok (xcmd, consumed) ->
+                        match xcmd with
+                        | XCommandConnected _ ->
+                            //TODO check server protocol version
+                            initialConnectionTsc.SetResult(this)
+                            reader.AdvanceTo consumed
```
Lanayx: Common consumed logic can now be extracted once for all cases

src/Pulsar.Client/Internal/ClientCnx.fs
```diff
@@ -212,73 +209,75 @@ type ClientCnx (broker: Broker,
...
-                        reader.AdvanceTo(consumed)
-                    | XCommandGetTopicsOfNamespaceResponse (cmd, consumed) ->
-                        let result = TopicsOfNamespace { Topics = List.ofSeq cmd.Topics }
-                        handleRespone %cmd.RequestId result reader consumed
-                    | XCommandError (cmd, consumed) ->
-                        Log.Logger.LogError("Error: {0}. Message: {1}", cmd.Error, cmd.Message)
-                        let result = Error
-                        handleRespone %cmd.RequestId result reader consumed
-                    | IncompleteCommand ->
+                    | Result.Error IncompleteCommand ->
```
Lanayx: No need to handle errors before Ok results, they happen much less often

src/Pulsar.Client/Internal/ClientCnx.fs
vilinski: you are completely right, the problem is only the `else` 50 lines later is difficult to find. Imaginary perf microoptimisations vs readability.  :)
At least I would extract a function instead, and then do it right.
I would reverse it again for you if you like the splitting DU idea generally, because I'm in doubt now whether here the right thing will be done.

So idea was just to make C#-ly `PulsarCommand` matchable with F#, or something else?

src/Pulsar.Client/Internal/ClientCnx.fs
vilinski: There's another now. But thats exactly the case where you probably don't need an exception?

src/Pulsar.Client/Internal/ClientCnx.fs
vilinski: right now I'm not that involved to know what the common logic here is. Maybe it is more suitable for another PR

src/Pulsar.Client/Internal/ClientCnx.fs
vilinski: see my first comment

src/Pulsar.Client/Internal/ClientCnx.fs
Lanayx: My comment was about the comparison line, not about exception =)

src/Pulsar.Client/Internal/ClientCnx.fs
Lanayx: I mean `reader.AdvanceTo consumed` line which can be extracted for all success cases to the bottom of match

src/Pulsar.Client/Internal/ClientCnx.fs
Lanayx: Yeah, the main idea is in 90% cases you don't need to find this else after 50 lines, because you want to understand the general algorithm. When rare cases are on the top it's more difficult to understand what's happening. 
Generally the split makes sense for me to extract `reader.AdvanceTo consumed` code from Ok match cases, without it there is not much use.

Issue #52, opened 2019-08-12
Implement deadLetter logic


Lanayx:
[Source](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L136-L140)

PR #73, opened 2019-11-14
[#52] Support dead letters

mprokushenkov:


src/Pulsar.Client/Internal/ConsumerImpl.fs
```diff
@@ -355,9 +381,13 @@ type ConsumerImpl internal (consumerConfig: ConsumerConfiguration, clientConfig:
                                 Log.Logger.LogDebug("{0} Ignoring message from before the startMessageId: {1}", prefix, startMessageId)
                             else
                                 let message = { rawMessage with MessageId = msgId } |> decompress
+
                                 if not hasWaitingChannel then
                                     incomingMessages.Enqueue(message)
+                                    storeDeadLetter message
```
Lanayx: storeDeadLetter in both branches, can it be moved to one level higher?

src/Pulsar.Client/Internal/ConsumerImpl.fs
```diff
@@ -412,8 +444,9 @@ type ConsumerImpl internal (consumerConfig: ConsumerConfiguration, clientConfig:
                             let messagesFromQueue = removeExpiredMessagesFromQueue(messageIds);
                             let batches = messageIds |> Seq.chunkBySize MAX_REDELIVER_UNACKNOWLEDGED
                             for batch in batches do
+                                let batch' = batch |> Array.filter (fun messageId -> processDeadLetters messageId |> not)
```
Lanayx: can you please rename batch' to nonDeadBatch (if I understand correctly)

tests/IntegrationTests/DeadLetters.fs
```diff
@@ -0,0 +1,215 @@
...
+                Task.Run(fun () ->
+                    task {
+                        do! consumeMessages dlqConsumer numberOfMessages dlqConsumerName
+                    }:> Task)
+
+            let tasks =
+                [|
+                    producerTask;
+                    consumerTask;
+                    Task.Delay(TimeSpan.FromSeconds(8.0))
```
Lanayx: 8 seconds is too much for test, can you reduce it to half of a second or less?

tests/IntegrationTests/DeadLetters.fs
```diff
@@ -0,0 +1,215 @@
...
+            let dlqConsumerTask =
+                Task.Run(fun () ->
+                    task {
+                        do! consumeMessages dlqConsumer numberOfMessages dlqConsumerName
+                    }:> Task)
+
+            let tasks =
+                [|
+                    producerTask;
+                    consumerTask;
```
Lanayx: semicolons are not necessary

tests/IntegrationTests/DeadLetters.fs
```diff
@@ -0,0 +1,215 @@
...
+    let numberOfMessages = 50
+
+    let logTestStart testDescription = Log.Debug(sprintf "Started '%s'" testDescription)
+    let logTestEnd testDescription = Log.Debug(sprintf "Finished '%s'" testDescription)
+
+    let buildProducer(producerName, topicName) =
+        ProducerBuilder(client)
+            .ProducerName(producerName)
+            .Topic(topicName)
+            .EnableBatching(false)
```
Lanayx: batching should be parametrized, we should check for both batched and non-batched cases

tests/IntegrationTests/DeadLetters.fs
```diff
@@ -0,0 +1,215 @@
...
+                Task.Run(fun () ->
+                    task {
+                        do! consumeMessages dlqConsumer numberOfMessages dlqConsumerName
+                    }:> Task)
+
+            let tasks =
+                [|
+                    producerTask;
+                    consumerTask;
+                    Task.Delay(TimeSpan.FromSeconds(8.0))
```
Lanayx: Too long =)

src/Pulsar.Client/Internal/ConsumerImpl.fs
```diff
@@ -161,6 +163,28 @@ type ConsumerImpl internal (consumerConfig: ConsumerConfiguration, clientConfig:
         | Some startMsgId ->
             subscriptionMode = SubscriptionMode.NonDurable && startMsgId.LedgerId = msgId.LedgerId && startMsgId.EntryId = msgId.EntryId
 
+    let clearDeadLetters() =
+        deadLettersProcessor
+        |> Option.iter (fun p -> p.ClearMessages())
```
Lanayx: Option.iter doesn't look good for me, can we implement DISABLED version to make the code cleaner?

src/Pulsar.Client/Api/DeadLetters.fs
```diff
@@ -0,0 +1,63 @@
...
+                store.[id].Add(message)
+
+        member __.RemoveMessage messageId =
+            store.Remove(messageId) |> ignore
+
+        member this.ProcessMessages messageId =
+            if  store.ContainsKey messageId then
+
+                store.[messageId]
+                |> List.ofSeq
```
Lanayx: Why is this cast needed?

src/Pulsar.Client/Api/DeadLetters.fs
```diff
@@ -0,0 +1,63 @@
...
+        let topicName =
+            if String.IsNullOrEmpty(policy.DeadLetterTopic) |> not then
+                policy.DeadLetterTopic
+            else
+                (sprintf "%s-%s-DLQ" (getTopicName()) (getSubscriptionNameName()))
+
+        createProducer topicName
+    )
+
+    let sendAndForgetAsync (builder : MessageBuilder) =
```
Lanayx: This is not sendAndForget, but send and ignore result =) Should we rename function or change called method?

tests/IntegrationTests/DeadLetters.fs
Lanayx: I suggest reverting logic by removing this statetement, since EnableBatching is true by default. To avoid chain .EnableBatching(false).EnableBatching(true)

src/Pulsar.Client/Api/DeadLetters.fs
Lanayx: This method SendAsync also can throw an exception

src/Pulsar.Client/Internal/ConsumerImpl.fs
mprokushenkov: Done.

src/Pulsar.Client/Internal/ConsumerImpl.fs
mprokushenkov: Done.

tests/IntegrationTests/DeadLetters.fs
mprokushenkov: Done.

src/Pulsar.Client/Api/DeadLetters.fs
mprokushenkov: It's unnecessary cast. Removed.

mprokushenkov: Please inspect `ProducerImpl.SendAndForgetAsync`. Body of this method contains similar code, but method named  `SendAndForgetAsync`.

```
member this.SendAndForgetAsync (message: byte[]) =
            task {
                connectionHandler.CheckIfActive()
                let! _ = this.SendMessage (MessageBuilder(message))
                return ()
            }
```

src/Pulsar.Client/Api/DeadLetters.fs
mprokushenkov: Please inspect `ProducerImpl.SendAndForgetAsync`. Body of this method contains similar code, but method named  `SendAndForgetAsync`.

```
member this.SendAndForgetAsync (message: byte[]) =
            task {
                connectionHandler.CheckIfActive()
                let! _ = this.SendMessage (MessageBuilder(message))
                return ()
            }
```

tests/IntegrationTests/DeadLetters.fs
mprokushenkov: Done.

src/Pulsar.Client/Api/DeadLetters.fs
Lanayx: Yes, the idea here is that SendMessage returns Task, which is not awaited for SendAndForgetAsync, so any error will be ignored. In your case the result is not ignored but awaited (even though ignored later), so exception could happen there

src/Pulsar.Client/Internal/ConsumerImpl.fs
mprokushenkov: Done!

src/Pulsar.Client/Api/DeadLetters.fs
mprokushenkov: Aaaa... Ok. May you suggest how to rewrite this method for task and possible exceptions ignorance?

tests/IntegrationTests/DeadLetters.fs
mprokushenkov: I decreased the interval.

tests/IntegrationTests/DeadLetters.fs
mprokushenkov: I decreased the interval.

src/Pulsar.Client/Internal/DeadLetters.fs
```diff
@@ -0,0 +1,68 @@
...
+            else
+                (sprintf "%s-%s-DLQ" (getTopicName()) (getSubscriptionNameName()))
+
+        createProducer topicName
+    )
+
+    let sendAndForgetAsync (builder : MessageBuilder) =
+        task {
+            let! p = producer.Value
+            p.SendAsync(builder) |> ignore
```
Lanayx: In Java version, they don't ignore sending error, but return false on failure, can we stick to that solution?

src/Pulsar.Client/Internal/ConsumerImpl.fs
```diff
@@ -396,6 +414,8 @@ type ConsumerImpl internal (consumerConfig: ConsumerConfiguration, clientConfig:
                         | _ ->
                             do! sendAcknowledge messageId ackType
                             Log.Logger.LogDebug("{0} acknowledged message - {1}, acktype {2}", prefix, messageId, ackType)
+
+                        removeDeadLetter messageId
```
Lanayx: can it be moved inside sendAcknowledge?

src/Pulsar.Client/Api/ConsumerBuilder.fs
```diff
@@ -101,6 +104,25 @@ type ConsumerBuilder private (client: PulsarClient, config: ConsumerConfiguratio
...
+    member __.DeadLettersPolicy (policy: DeadLettersPolicy) =
+
+        let ackTimeoutTickTime =
+            if config.AckTimeoutTickTime = TimeSpan.Zero
+            then TimeSpan.FromMilliseconds(DEFAULT_ACK_TIMEOUT_MILLIS_FOR_DEAD_LETTER)
+            else config.AckTimeoutTickTime
+
+        let getTopicName() = config.Topic.ToString()
+        let getSubscriptionName() = config.SubscriptionName
+        let createProducer deadLetterTopic = ProducerBuilder(client).Topic(deadLetterTopic).CreateAsync()
```
Lanayx: MaxRedeliveryCount is missing

src/Pulsar.Client/Internal/DeadLetters.fs
```diff
@@ -0,0 +1,78 @@
...
+        member __.RemoveMessage messageId =
+            store.Remove(messageId) |> ignore
+
+        member __.ProcessMessages messageId =
+            if  store.ContainsKey messageId then
+                store.[messageId]
+                |> Seq.map (fun m -> MessageBuilder(m.Payload, %m.MessageKey, m.Properties))
+                |> Seq.map (fun builder -> sendMessageAsync builder messageId)
+                |> Async.Parallel
+                |> Async.RunSynchronously
```
Lanayx: Two issues
1) Java version doesn't send in parallel, let's stick to it
2) RunSynchronously blocks thread, please make ProcessMessages async

mprokushenkov: I think code should block this thread because code should wait send result and return true or false...? All messages that are bundled under same message id (batch) should be sent. If one of sending attempt throws an exception function returns false. So, I think it's necessarily wait send result.

Lanayx: I agree that result is needed, still there is no need to block thread for it :) Just make the method async and await it on the caller side

src/Pulsar.Client/Api/ConsumerBuilder.fs
mprokushenkov: `MaxRedeliveryCount ` is missing where?

Lanayx: Sorry, mixed up :)

src/Pulsar.Client/Internal/ConsumerImpl.fs
mprokushenkov: Ok. I did move it.

src/Pulsar.Client/Internal/DeadLetters.fs
mprokushenkov: Ok. Now dead letters processed in series and await moved on the caller side.

Lanayx: Refactored it, please take a look what I wanted ;)

mprokushenkov: Ok. I saw it. Now the idea is clear for me. I think there is small difference. As result you moved async await farther into MailboxAgent branch. But end effect will be the same I think.

src/Pulsar.Client/Internal/DeadLetters.fs
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+    let sendMessage (builder : MessageBuilder) messageId  =
+        task {
+            let! p = producer.Value
+            let! _ = p.SendAsync(builder)
+            return ()
+        }
+
+    let toStoreFormat (messageId : MessageId) =
```
Lanayx: Why is this method needed? Don't see such cast in Java code

mprokushenkov: This method converts MessageId into key suitable for storing single and batch messages.
Notice. I does not copy java code line by line, because our project in other language. And original java code sometime looks unwanted tangled.

PR #75, opened 2019-12-25
[#77] Implement scheduled message delivery

mprokushenkov:


src/Pulsar.Client/Internal/ProducerImpl.fs
```diff
@@ -131,6 +131,9 @@ type internal ProducerImpl private (producerConfig: ProducerConfiguration, clien
                 metadata.Properties.Add(KeyValue(Key = property.Key, Value = property.Value))
         if numMessagesInBatch.IsSome then
             metadata.NumMessagesInBatch <- numMessagesInBatch.Value
+
+        message.DeliverAfter |> Option.iter (fun ts -> metadata.DeliverAtTime <- ts)
```
Lanayx: Can you please simplify it to if as few lines about (because it's a hot path and we should avoid additional method invocations). Also note, that we are setting DeliverAtTime, not DeliverAfter

PR #79, opened 2020-02-28
Converting to Github Actions CI

### LINKED ISSUE #74: Fix CI build
Either Travis should be fixed, or CI moved to Github actions

Szer:
fixes https://github.com/fsharplang-ru/pulsar-client-dotnet/issues/74

Added 2 workflows (for different badges):
- `dotnetcore-windows.yml`
- `dotnetcore-ubuntu.yml`

As part of this PR also included:
- FAKE removal (nobody needs it, it's just redundant wrapper on top of dotnet CLI)
- Other CI removal (Azure-pipelines and Travis)
- Updated Readme


Szer:
Fixed flaky tests by adding optional parameter to both Trackers to control when `TickTime` happens.

If this argument is not presented (which is normal), Tracker will create the good old Timer without any overhead.
If passed (unit test case) we could come up with our own object to `tick` (and this object is `ManualInvokeScheduler`)

Issue #57, opened 2019-08-27
Support interceptors for different operations


Lanayx:
[Consumer interceptor](https://github.com/apache/pulsar/blob/master/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerInterceptor.java)
[Producer interceptor](https://github.com/apache/pulsar/blob/master/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ProducerInterceptor.java)

PR #84, opened 2020-03-13
[#57] Interceptors implementation

RahimovIR:


src/Pulsar.Client/Internal/ConsumerImpl.fs
```diff
@@ -145,7 +154,7 @@ type internal ConsumerImpl internal (consumerConfig: ConsumerConfiguration, clie
                 Log.Logger.LogDebug("{0} update PrevBatchCumulativelyAcked", prefix)
                 batchAcker.PrevBatchCumulativelyAcked <- true
             | _ ->
-                ()
+                interceptors.OnAcknowledge(this, msgId, null)
```
Lanayx, line 224: Can you please fix by moving it out of match statement and splitting into OnAcknowledge and OnAcknowledgeCumulative based on ack type?

src/Pulsar.Client/Internal/ConsumerImpl.fs
```diff
@@ -788,45 +798,59 @@ type internal ConsumerImpl internal (consumerConfig: ConsumerConfiguration, clie
...
                 mb.Post(Acknowledge(msgId, AckType.Individual))
             }
             
         member this.AcknowledgeAsync (msgs: Messages) =
             task {
-                connectionHandler.CheckIfActive()
+                let exn = connectionHandler.CheckIfActive()
+                if not (isNull exn) then 
+                    for msg in msgs do
+                        interceptors.OnAcknowledge(this, msg.MessageId, null)
```
Lanayx: shouldn't be an exn there instead of null?

src/Pulsar.Client/Internal/ConsumerImpl.fs
RahimovIR: Completed

src/Pulsar.Client/Internal/ConsumerImpl.fs
RahimovIR: Fixed

Lanayx (approved):

PR #109, opened 2020-06-24
Avro support should be able to use Avro generated C# classes

richardwilko:
This lets the avro schema support use the generated avro c# classes


src/Pulsar.Client/Schema/AvroSchema.fs
```diff
@@ -6,22 +6,38 @@ open Avro.Generic
...
-    let avroReader = ReflectReader<'T>(avroSchema, avroSchema)
     
-    new () = AvroSchema(typeof<'T>.GetSchema())
+    new () =
+         let tpe = typeof<'T>
+         if typeof<ISpecificRecord>.IsAssignableFrom(tpe) then
+            let avroSchema = downcast tpe.GetField("_SCHEMA").GetValue(null) :Schema
+            let avroWriter = SpecificDatumWriter<'T>(avroSchema)
+            // note afaik for schema evolutions, the writerSchema (first param) should come from the topic
+            // (i.e the schema the value was written in)
```
Lanayx: Right! Therefore can you please update **GetSpecificSchema** method, since it takes the schema from the topic as the first parameter

richardwilko: ah, okay, i was wondering about that

richardwilko: something like this:
```f#
    override this.GetSpecificSchema stringSchema =
        let writtenSchema = Schema.Parse(stringSchema)
        if avroReader :? SpecificDatumReader<'T> then
            AvroSchema(schema, SpecificDatumReader(writtenSchema, schema), avroWriter) :> ISchema<'T>
        else
            AvroSchema(stringSchema) :> ISchema<'T>
```

Lanayx: looks good to me, the only thing to change is to put the parse logic inside if rather than outside

richardwilko: though wouldn't the reflection based version also need the same thing?
```f#
    override this.GetSpecificSchema stringSchema =
        let writtenSchema = Schema.Parse(stringSchema)
        if avroReader :? SpecificDatumReader<'T> then
            AvroSchema(schema, SpecificDatumReader(writtenSchema, schema), avroWriter) :> ISchema<'T>
        else
            AvroSchema(schema, ReflectReader<'T>(writtenSchema, schema), avroWriter) :> ISchema<'T>
```

Lanayx: Hm... Good catch! I think you are right

Lanayx:
@richardwilko the integraiton test fails, looks like SimpleRecord is not compatible with [SimpleRecord2](QSPublisher), can you please try constructing second SimpleRecord using dynamic assembly in runtime? ( similar to this response https://stackoverflow.com/questions/929349/is-there-a-way-to-build-a-new-type-during-runtime or https://stackoverflow.com/questions/3862226/how-to-dynamically-create-a-class/3862241#3862241)

richardwilko:
@Lanayx Will the environment let you define a class with the same name and namespace as one that already exists though?  

Feels like updating AvroSchemaGenerator to support name [aliases ](https://avro.apache.org/docs/current/spec.html#Aliases) with another annotation like `[DefaultValue("200")]` might be easier?

However the easiest way I've found to test schema changes for avro is actually using GenericRecord, though would require a bit more work to support those as well.

Ironically, the actual Java version of avro does not seem to care if the name changes, and this code would pass :)

Lanayx:
@richardwilko 
Yes, the environment will allow to do that if the type is defined in another (dynamically created in our case) assembly.
As for generic records, I didn't find good usecases at that moment, since most people would be happy to use static types, so skipped this effort.
As for aliases - this is a great idea, can you please create an issue in AvroSchemaGenerator repo to add support for them. Looks like a custom Aliases attribute should be defined in that library.

Lanayx:
@richardwilko it looks like the Alias support has been added, so can you try updating the integration test with the SimpleRecord2 in failing integration test with the attribute? To update the packages you'll need to run `dotnet paket update` command from the root folder

src/Pulsar.Client/Schema/AvroSchema.fs
```diff
@@ -35,7 +49,16 @@ type internal AvroSchema<'T> private (stringSchema) =
         use stream = new MemoryStream(bytes)
         avroReader.Read(Unchecked.defaultof<'T>, BinaryDecoder(stream))        
     override this.GetSpecificSchema stringSchema =
-        AvroSchema(stringSchema) :> ISchema<'T>
+        let writtenSchema = Schema.Parse(stringSchema)
+        if avroReader :? SpecificDatumReader<'T> then
+            AvroSchema(schema, SpecificDatumReader(writtenSchema, schema), avroWriter) :> ISchema<'T>
+        else
+            //Avro doesnt figure that the written classname might be different from the reader classname
+            //Seems like it might be a bug in ReflectReader, but this works around that
```
richardwilko, line 57: Not sure if found an avro bug here or not.  They do talk about manually registering in the docs, but not in this context: https://github.com/apache/avro/tree/master/lang/csharp/src/apache/main/Reflect#manual-registration

Lanayx: I see, in that case you can try raising an issue in their Jira tracker if this is an expected behaviour. As for us it looks this workaround is needed anyway, so merging the PR, big thanks!

Lanayx (approved):

src/Pulsar.Client/Schema/AvroSchema.fs
richardwilko: fyi created [AVRO-2873](https://issues.apache.org/jira/browse/AVRO-2873) in the Apache JIRA,

PR #116, opened 2020-08-19
Change image to apachepulsar/pulsar:2.6.0

RahimovIR:


tests/compose/standalone/docker-compose.yml
```diff
@@ -6,13 +6,13 @@ version: '2.1'
 services:
   standalone:
     hostname: standalone
-    image: apachepulsar/pulsar-test-latest-version:latest
+    image: apachepulsar/pulsar:2.6.0
```
Lanayx: Let's change the tag to latest so we won't need to update it each time (in other places as well)

.github/workflows/dotnetcore-ubuntu.yml
```diff
@@ -47,7 +47,6 @@ jobs:
         cd tests/compose/standalone
         docker-compose ps
         docker-compose logs init
-        docker-compose logs initTls
```
Lanayx, line 50: We'll need a separate cluster, since in one cluster we'll need maxMessageSize to be default and in another one increased.

Lanayx (approved):

PR #133, opened 2020-12-14
Failing test for Avro with string prop > 256 characters

JarrodJ83:


Lanayx:
@JarrodJ83 The problem is in Avro itself, there are two PRs that resolve the issue, but neither of them is merged yet https://github.com/apache/avro/pull/957 and https://github.com/apache/avro/pull/1013

PR #142, opened 2021-01-22
Replicate "Duplicate schema name" exception

JarrodJ83:
This duplicates an exception I am receiving when trying to reuses classes in my schemas. In this case the re-used class is `AvroSchemaTest`. It seems to be local to Pulsar.Client as calling `typeof(UsesSharedSchema).GetSchema()` succeeds.


JarrodJ83:
Closed because issue is in AvroSchemaGenerator.

PR #165, opened 2021-06-10
Oauth2 support

fjod:
should be good enough for review!


src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+//Gets a well-known metadata URL for the given OAuth issuer URL.
+//https://tools.ietf.org/id/draft-ietf-oauth-discovery-08.html#ASConfig
+let getWellKnownMetadataUrl (issuerUrl: Uri) : Uri =
+    Uri(issuerUrl.AbsoluteUri + ".well-known/openid-configuration")
+
+let getMetadata (issuerUrl: Uri) : Async<Metadata> =
+    async {
+        use client = new HttpClient()
```
Lanayx: Could you make use of HttpClientFactory instead?

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,117 @@
...
+        return JsonSerializer.Deserialize<Metadata> response        
+    }
+
+let createClient issuerUrl =
+    let data = getMetadata issuerUrl |> Async.RunSynchronously  
+    TokenClient(Uri(data.TokenEndpoint))
+
+let openAndDeserializeCreds uri =
+    let text = System.IO.File.ReadAllText uri    
+    JsonSerializer.Deserialize<Credentials>(text)
```
Lanayx: Can we read the file into a stream and deserialize from the stream?

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,117 @@
...
+        IssuerUrl : string 
+    }
+
+//Gets a well-known metadata URL for the given OAuth issuer URL.
+//https://tools.ietf.org/id/draft-ietf-oauth-discovery-08.html#ASConfig
+let getWellKnownMetadataUrl (issuerUrl: Uri) : Uri =
+    Uri(issuerUrl.AbsoluteUri + ".well-known/openid-configuration")
+
+let getMetadata (issuerUrl: Uri) : Async<Metadata> =
+    async {
```
Lanayx: We use task builder instead of async in this library

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,117 @@
...
+        use client = new HttpClient()
+        let metadataDataUrl = getWellKnownMetadataUrl issuerUrl        
+        let! response = client.GetStringAsync metadataDataUrl |> Async.AwaitTask
+        return JsonSerializer.Deserialize<Metadata> response
```
Lanayx: Can we deserialize right from the response stream without converting to string first?

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,117 @@
...
+    let credentials = openAndDeserializeCreds(credentials.LocalPath) 
+    let mutable token : Option<TokenResult * DateTime> = None   
+
+    //https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2
+    let isTokenExpiredOrEmpty() : bool =
+        match token with
+        | Some v ->
+            let tokenDuration = TimeSpan.FromSeconds(float (fst v).ExpiresIn)
+            let tokenExpiration = (snd v).Add tokenDuration
+            DateTime.Now > tokenExpiration
```
Lanayx: DateTime.UtcNow should be used (I suppose that token ExpiresIn is in UTC)

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,117 @@
...
+
+            match newToken with
+            | Result (v, d) ->
+                token <- Some(v, d)
+                returnTokenAsProvider()
+            | OauthError e ->
+                raise (
+                    TokenExchangeException
+                        (                        
+                        e.Error
```
Lanayx: Can you use F# interpolation here?

src/Pulsar.Client/Auth/TokenClient.fs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+type TokenExchangeResult =
+    | Result of TokenResult * DateTime
+    | OauthError of TokenError
+    | OtherError of string
+    
+
+let Exchange (uri:Uri,clientId:string, clientSecret:string, audience:Uri) : TokenExchangeResult =
+     async{
```
Lanayx: TaskBuilder here as well. Also please change function name to lowercase and make parameters to carried form

src/Pulsar.Client/Auth/TokenClient.fs
```diff
@@ -0,0 +1,82 @@
...
+             use client = new HttpClient()            
+             let request = new HttpRequestMessage(HttpMethod.Post,uri)
+           
+             request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue("application/json"))             
+             request.Headers.Add("User-Agent","Pulsar-Java-v2.7.1")
+             
+             let body = [
+                         KeyValuePair("grant_type","client_credentials");
```
Lanayx: Semicolons are not needed

src/Pulsar.Client/Auth/TokenClient.fs
```diff
@@ -0,0 +1,82 @@
...
+    | OtherError of string
+    
+
+let Exchange (uri:Uri,clientId:string, clientSecret:string, audience:Uri) : TokenExchangeResult =
+     async{
+             use client = new HttpClient()            
+             let request = new HttpRequestMessage(HttpMethod.Post,uri)
+           
+             request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue("application/json"))             
+             request.Headers.Add("User-Agent","Pulsar-Java-v2.7.1")
```
Lanayx: We are not Java :) And I think we can skip version here

src/Pulsar.Client/Auth/TokenClient.fs
```diff
@@ -0,0 +1,82 @@
...
+             let! resultContent = response.Content.ReadAsStringAsync() |> Async.AwaitTask
+             
+             match response.StatusCode with
+                | HttpStatusCode.OK ->                   
+                    let result =  JsonSerializer.Deserialize<TokenResult>(resultContent)
+                    return TokenExchangeResult.Result (result, DateTime.Now)
+                | HttpStatusCode.BadRequest
+                | HttpStatusCode.Unauthorized ->
+                    let! resultContent = response.Content.ReadAsStringAsync() |> Async.AwaitTask
+                    let result =  JsonSerializer.Deserialize<TokenError>(resultContent)
```
Lanayx: Same comments about deserialization

src/Pulsar.Client/Auth/TokenClient.fs
```diff
@@ -0,0 +1,82 @@
...
+                        ]           
+           
+             request.Content <- new FormUrlEncodedContent(body)
+            
+             let! response = client.SendAsync request |> Async.AwaitTask
+             let! resultContent = response.Content.ReadAsStringAsync() |> Async.AwaitTask
+             
+             match response.StatusCode with
+                | HttpStatusCode.OK ->                   
+                    let result =  JsonSerializer.Deserialize<TokenResult>(resultContent)
```
Lanayx: Same comments about deserialization

src/Pulsar.Client/Auth/TokenClient.fs
```diff
@@ -0,0 +1,82 @@
...
+                    return TokenExchangeResult.Result (result, DateTime.Now)
+                | HttpStatusCode.BadRequest
+                | HttpStatusCode.Unauthorized ->
+                    let! resultContent = response.Content.ReadAsStringAsync() |> Async.AwaitTask
+                    let result =  JsonSerializer.Deserialize<TokenError>(resultContent)
+                    return TokenExchangeResult.OauthError result
+                | _ -> return TokenExchangeResult.OtherError ("Failed to perform request to oauth server "
+                                                              + response.StatusCode.ToString()
+                                                              + " " +    response.ReasonPhrase)   
+          } |> Async.RunSynchronously
```
Lanayx: We shouldn't block threads

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,117 @@
...
+
+let openAndDeserializeCreds uri =
+    let text = System.IO.File.ReadAllText uri    
+    JsonSerializer.Deserialize<Credentials>(text)    
+
+type AuthenticationOauth2(issuerUrl: Uri, credentials: Uri, audience: Uri) =
+    inherit Authentication()
+    let tokenClient  = createClient issuerUrl
+    let credentials = openAndDeserializeCreds(credentials.LocalPath) 
+    let mutable token : Option<TokenResult * DateTime> = None
```
Lanayx: Can we avoid mutable variable?

src/Pulsar.Client/Auth/TokenClient.fs
```diff
@@ -0,0 +1,82 @@
...
+        RefreshToken:string
+        
+        [<JsonPropertyName("expires_in")>]
+        ExpiresIn:int 
+    }
+
+type TokenExchangeResult =
+    | Result of TokenResult * DateTime
+    | OauthError of TokenError
+    | OtherError of string
```
Lanayx: Two types of error look suspicious to me

examples/Oauth2Files/standalone.conf
```diff
@@ -0,0 +1,979 @@
...
+
+# Enable or disable topic level policies, topic level policies depends on the system topic
+# Please enable the system topic first.
+topicLevelPoliciesEnabled=false
+
+# If a topic remains fenced for this number of seconds, it will be closed forcefully.
+# If it is set to 0 or a negative number, the fenced topic will not be closed.
+topicFencingTimeoutSeconds=0
+
+### --- Authentication --- ###
```
Lanayx, line 1: Can we remove everything in this file exception Authentication section?

examples/Oauth2Files/docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+﻿version: "3.9"
```
Lanayx: This file can be removed

examples/Oauth2Files/client.conf
```diff
@@ -0,0 +1,69 @@
+#
```
Lanayx: This file can be removed

examples/Oauth2Files/Dockerfile
```diff
@@ -0,0 +1,13 @@
+﻿FROM apachepulsar/pulsar-standalone:2.7.1
```
Lanayx: This file can be removed

examples/Oauth2Files/Auth0response
```diff
@@ -0,0 +1 @@
+{"access_token":"eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6IkM0OWg1NHJsWWY2N2NZM1psbUlOMSJ9.eyJodHRwczovL3B1bHNhci5jb20vdG9rZW5BdXRoQ2xhaW0iOiJsb3JkX2tlayIsImlzcyI6Imh0dHBzOi8vcHVsc2FyLXNhbXBsZS51cy5hdXRoMC5jb20vIiwic3ViIjoia0MydDhvRjBrWWJRaXVqNUZRR0hYSFp6ZXIyRkI0NnRAY2xpZW50cyIsImF1ZCI6Imh0dHBzOi8vcHVsc2FyLXNhbXBsZS51cy5hdXRoMC5jb20vYXBpL3YyLyIsImlhdCI6MTYyMzMwODQ3MCwiZXhwIjoxNjIzMzk0ODcwLCJhenAiOiJrQzJ0OG9GMGtZYlFpdWo1RlFHSFhIWnplcjJGQjQ2dCIsImd0eSI6ImNsaWVudC1jcmVkZW50aWFscyJ9.KmbwZX88qpCRYCA3UvjrhGJqKsQphWh3kyG2ZIgGPLefLWO2PeloaVoPAqZbA7jMUrxj1BiZtA0gdlDO4oNfqVdzvVXWEwoFBrSqZHmvohJFoGAueO9j0fFjvwzo5f6d_G2KctTpneYtU-bbMdplXoG3y3iFtjpmC1nd5sh6Q9jsc9FNF8K_acLmfFaKOsiF7m3FPt6tjcYkoLHMNu5bsU_ES1DdF0bzbndsQ6UwY1u81IaInJODLvSbTLVCGIRx8mSuIaUMVuCLQr17OWamwdVaqqKT5SczSlz7TTfDldqJYvZEWNfmfAsMfP-8kVsq2LkB_WTj-obk6wP46j49Tw","expires_in":86400,"token_type":"Bearer"}
```
Lanayx: This file can be removed

examples/Oauth2Files/credentials_file.json
```diff
@@ -0,0 +1,7 @@
+﻿{
```
Lanayx: This file can be removed

examples/Oauth2Files/credentials_file.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "client_credentials",
+  "client_id": "kC2t8oF0kYbQiuj5FQGHXHZzer2FB46t",
+  "client_secret": "wOSnirygS3E-dUKSDnEdyclwvRa4l2jFPlvAdJMa3U1_zCWrjcvng0uYKXF2XZn6",
+  "client_email": "test@developer.gserviceaccount.com",
+  "issuer_url": "https://pulsar-sample.us.auth0.com/oauth/token"
```
Lanayx: What is the rule - when oauth/token should be on the end and when it shouldn't?

src/Pulsar.Client/Auth/TokenClient.fs
```diff
@@ -0,0 +1,82 @@
...
+                    let! resultContent = response.Content.ReadAsStringAsync() |> Async.AwaitTask
+                    let result =  JsonSerializer.Deserialize<TokenError>(resultContent)
+                    return TokenExchangeResult.OauthError result
+                | _ -> return TokenExchangeResult.OtherError ("Failed to perform request to oauth server "
+                                                              + response.StatusCode.ToString()
+                                                              + " " +    response.ReasonPhrase)   
+          } |> Async.RunSynchronously
+     
+type TokenClient(tokenUrl : Uri) =
+      member x.exchangeClientCredentials(clientId:string, clientSecret:string, audience:Uri)=
```
Lanayx: Method should be from uppercase, this used instead of x, also please add whitespaces inside the body

Lanayx (changes requested):

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
fjod: It would need to add dependency on System.Net.Http and DI like [here](https://docs.microsoft.com/en-us/previous-versions/aspnet/hh995280(v=vs.108)?redirectedfrom=MSDN),  [here](https://www.stevejgordon.co.uk/introduction-to-httpclientfactory-aspnetcore) or [here](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.ihttpclientfactory?view=dotnet-plat-ext-5.0). On the other hand, it uses one request to get metadata on startup and one request to get/renew token; so I'm not sure if pros of using factory will outweigh cons of adding another nuget.

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
fjod: There is only one overload which accepts stream, and it's async. So I come with

let openAndDeserializeCreds uri =
    let q = task{
        use fs = new FileStream(uri, FileMode.Open)
        let! temp = JsonSerializer.DeserializeAsync<Credentials>(fs)
        return temp
    }
    q.GetAwaiter().GetResult()

Not sure if it's better (we are parsing file of 5 lines)

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
fjod: I guess if we use like list of tokens with length of 1. If list is emtpy, then it's time to fetch. If not - then proceed to check expiration. Can not think of something better.

examples/Oauth2Files/credentials_file.json
fjod: Examples wont work without it

examples/Oauth2Files/credentials_file.json
fjod: They should be on the end for client auth and should not for broker auth

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
Lanayx: Right, it's not better :) You should turn it into 
```
let openAndDeserializeCreds uri =
  task{
    use fs = new FileStream(uri, FileMode.Open)
    let! temp = JsonSerializer.DeserializeAsync(fs)
    return temp
  }
```

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
Lanayx: Ok, in that case can you pleas pass this instance of httpclient to the TokenClient and reuse it on refresh requests.

examples/CsharpExamples/Oauth2.cs
```diff
@@ -0,0 +1,62 @@
...
+            var issuerUrl = new Uri("https://pulsar-sample.us.auth0.com");
+            var audience = new Uri("https://pulsar-sample.us.auth0.com/api/v2/");
+
+            var serviceUrl = "pulsar://localhost:6650";
+            var subscriptionName = "my-subscription";
+            var topicName =  $"my-topic-%{DateTime.Now.Ticks}";
+            
+            var client = await new PulsarClientBuilder()
+                .ServiceUrl(serviceUrl)
+                //.Authentication(AuthenticationFactory.oauth2(issuerUrl,fileUri,audience))  //try to comment out this line
```
Lanayx: We should use AuthenticationFactoryOAuth2.clientCredentials like in Java version

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,116 @@
...
+    task{
+        use fs = new FileStream(uri, FileMode.Open)
+        let! temp = JsonSerializer.DeserializeAsync<Credentials>(fs)
+        return temp
+    }
+    
+type AuthenticationOauth2(issuerUrl: Uri, credentials: Uri, audience: Uri) =
+    inherit Authentication()
+    let tokenClient  = (createClient issuerUrl).GetAwaiter().GetResult()
+    let credentials = openAndDeserializeCreds(credentials.LocalPath).GetAwaiter().GetResult()
```
Lanayx: openAndDeserializeCreds should be called on each refresh, not only once

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,116 @@
...
+type AuthenticationOauth2(issuerUrl: Uri, credentials: Uri, audience: Uri) =
+    inherit Authentication()
+    let tokenClient  = (createClient issuerUrl).GetAwaiter().GetResult()
+    let credentials = openAndDeserializeCreds(credentials.LocalPath).GetAwaiter().GetResult() 
+    let mutable token : Option<TokenResult * DateTime> = None   
+
+    //https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2
+    let isTokenExpiredOrEmpty() : bool =
+        match token with
+        | Some v ->
```
Lanayx: Can you please decompose v here rather than using fst and snd below

examples/Oauth2Files/credentials_file.json
fjod: After re-check it turned out that both client and broker can work without oauth/token on the end

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
fjod: wow I did not know I can do that xD

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,115 @@
...
+    }
+    
+let openAndDeserializeCreds uri =
+    task{
+        use fs = new FileStream(uri, FileMode.Open)
+        let! temp = JsonSerializer.DeserializeAsync<Credentials>(fs)
+        return temp
+    }
+    
+type AuthenticationOauth2(issuerUrl: Uri, credentials: Uri, audience: Uri) =
```
Lanayx: Audience is not Uri in Java version, can you please also rename credentials to privateKey (like in Java version)

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,115 @@
...
+            let tokenDuration = TimeSpan.FromSeconds(float res.ExpiresIn)
+            let tokenExpiration = exp.Add tokenDuration
+            DateTime.UtcNow > tokenExpiration
+        | _ -> true
+        
+    override this.GetAuthMethodName() = "token"
+    override this.GetAuthData() =
+        
+            let returnTokenAsProvider () =
+              AuthenticationDataToken(fun () -> (fst token.Value).AccessToken) :> AuthenticationDataProvider
```
Lanayx: Can we avoid using .Value on Option and fst? It would make sense for me to pass accessToken as a parameter

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,115 @@
...
+let openAndDeserializeCreds uri =
+    task{
+        use fs = new FileStream(uri, FileMode.Open)
+        let! temp = JsonSerializer.DeserializeAsync<Credentials>(fs)
+        return temp
+    }
+    
+type AuthenticationOauth2(issuerUrl: Uri, credentials: Uri, audience: Uri) =
+    inherit Authentication()
+    let tokenClient  = (createClient issuerUrl).GetAwaiter().GetResult()
```
Lanayx: Let's not await here, rather inside GetAuthData. Feild can be renamed to tokenClientTask

src/Pulsar.Client/Auth/TokenClient.fs
```diff
@@ -0,0 +1,86 @@
...
+                | HttpStatusCode.BadRequest
+                | HttpStatusCode.Unauthorized ->
+                    let! resultContent = response.Content.ReadAsStreamAsync() 
+                    let! result =  JsonSerializer.DeserializeAsync<TokenError>(resultContent)
+                    return TokenExchangeResult.OauthError result
+                | _ -> return TokenExchangeResult.HttpError $"Failed to perform request to oauth server {response.StatusCode.ToString()} {response.ReasonPhrase}"
+                                                                
+          } 
+     
+type TokenClient(tokenUrl : Uri) =
```
Lanayx: Can you please pass HttpClient instanse here and use it during exchange (instead of creating new HttpClient instances). It should also be used for metadata and disposed on TokenClient disposal (also implement IDisposable for TokenClient and parent AuthenticationOauth2 and Authentication, Authentication should be disposed at PulsarClient Close)

src/Pulsar.Client/Auth/AuthenticationOauth2.fs
```diff
@@ -0,0 +1,115 @@
...
+    let tokenClient  = (createClient issuerUrl).GetAwaiter().GetResult()     
+    let mutable token : Option<TokenResult * DateTime> = None  
+
+    //https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2
+    let isTokenExpiredOrEmpty() : bool =
+        match token with
+        | Some (res, exp) ->
+            let tokenDuration = TimeSpan.FromSeconds(float res.ExpiresIn)
+            let tokenExpiration = exp.Add tokenDuration
+            DateTime.UtcNow > tokenExpiration
```
Lanayx: Can you please return back to DateTime.Now, since in another place it is still DateTime.Now, and this value is not sent anywhere, just hold in memory and compared only here.

src/Pulsar.Client/Auth/TokenClient.fs
```diff
@@ -0,0 +1,87 @@
...
+             let request = new HttpRequestMessage(HttpMethod.Post,uri)
+           
+             request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue("application/json"))
+             request.Headers.Add("User-Agent", "Pulsar.Client")
+                         
+             let body = [
+                         KeyValuePair("grant_type","client_credentials")
+                         KeyValuePair("client_id",clientId)
+                         KeyValuePair("client_secret",clientSecret)
+                         KeyValuePair("audience",audience.ToString())
```
Lanayx: .ToString() is no longer needed

fjod: sorry about that  ¯\\\_(ツ)_/¯

PR #180, opened 2021-08-05
Feautre/175 support seek by function

Iezious:


src/Pulsar.Client/Common/DTO.fs
```diff
@@ -477,7 +477,7 @@ type internal PulsarResponseType =
 
 type internal ResultOrException<'T> = Result<'T, exn>
 
-type internal SeekData =
+type SeekData =
```
Lanayx: Can you please rename SeekData to SeekType

Lanayx (approved):

PR #182, opened 2021-08-25
Exposed redelivery count of the internal msg to the public contract.

kuskmen:
I didnt feel the need to write tests for that, but if you want I will add some. (Tested this locally and it works as expected in both scenarios - negatively acknowledging and acknowledge timeout.

```csharp
8/25/2021 9:38:51 AM Input: 1125, Redelivery count: 0
8/25/2021 9:38:56 AM Input: 1125, Redelivery count: 1
8/25/2021 9:39:01 AM Input: 1125, Redelivery count: 2
8/25/2021 9:39:06 AM Input: 1125, Redelivery count: 3
8/25/2021 9:39:11 AM Input: 1125, Redelivery count: 4
8/25/2021 9:39:16 AM Input: 1125, Redelivery count: 5
8/25/2021 9:39:21 AM Input: 1125, Redelivery count: 6
```

RedeliveryCount is already exposed in java client
https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java#L737

### DIFF
--- a/src/Pulsar.Client/Common/DTO.fs
+++ b/src/Pulsar.Client/Common/DTO.fs
@@ -260,6 +260,7 @@ type Message<'T> internal (messageId: MessageId, data: byte[], key: PartitionKey
                   properties: IReadOnlyDictionary<string, string>, encryptionCtx: EncryptionContext option,
                   schemaVersion: byte[], sequenceId: SequenceId, orderingKey: byte[], publishTime: TimeStamp,
                   eventTime: Nullable<TimeStamp>,
+                  redeliveryCount: uint32, 
                   getValue: unit -> 'T) =
     /// Get the unique message ID associated with this message.
     member this.MessageId = messageId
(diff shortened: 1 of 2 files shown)


src/Pulsar.Client/Common/DTO.fs
```diff
@@ -260,6 +260,7 @@ type Message<'T> internal (messageId: MessageId, data: byte[], key: PartitionKey
                   properties: IReadOnlyDictionary<string, string>, encryptionCtx: EncryptionContext option,
                   schemaVersion: byte[], sequenceId: SequenceId, orderingKey: byte[], publishTime: TimeStamp,
                   eventTime: Nullable<TimeStamp>,
+                  redeliveryCount: uint32,
```
Lanayx: Can you please expose it as int32 to conform Java version

kuskmen: @Lanayx, personally I think Java version is poorly written if the broker sends you uint, but okay I guess. 
To me, it does not make sense the statement `RedeliveryCount = -156`

Lanayx: Right, but it's not the only place with such discrepancy :)

PR #185, opened 2021-08-31
Add multi-version support for AutoConsumeSchema.

RobertIndie:
### Motivation
Currently, AutoConsumeSchema does not support multi-version schema. This will cause each message to be decoded with the latest schema, instead of getting the correct version schema based on the schemaVersion of the message. This eventually leads to unexpected errors when decoding data.

This PR adds multi-version support for AutoConsumeSchema. 

### Modification
* Set AutoConsumeSchema. SupportSchemaVersioning to true.
* Change the function `ISchema.GetSpecificSchema ` so that we can pass both `SchemaVersion` and `SchemaInfo` to the GenericAvroSchema.
* Override GenericAvroSchema.GetSpecificSchema so that we can use the correct schema version of GenericAvroSchema.

### DIFF
--- a/src/Pulsar.Client/Schema/AvroSchema.fs
+++ b/src/Pulsar.Client/Schema/AvroSchema.fs
@@ -49,8 +49,8 @@ type internal AvroSchema<'T> private (schema: Schema, avroReader: DatumReader<'T
     override this.Decode bytes =
         use stream = new MemoryStream(bytes)
         avroReader.Read(defaultValue, BinaryDecoder(stream))
-    override this.GetSpecificSchema stringSchema =
-        let writtenSchema = Schema.Parse(stringSchema)
+    override this.GetSpecificSchema schemaInfo _ =
+        let writtenSchema = Schema.Parse(schemaInfo.Schema |> Encoding.UTF8.GetString)
         if avroReader :? SpecificDatumReader<'T> then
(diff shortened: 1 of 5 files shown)


Lanayx:
Hi, thank you for the PR, can you please add an integration test for versioned multiconsume schema. [Example](https://github.com/fsharplang-ru/pulsar-client-dotnet/blob/develop/tests/IntegrationTests/Schema.fs#L366-L396)

src/Pulsar.Client/Schema/AvroSchema.fs
```diff
@@ -84,8 +87,11 @@ type internal GenericAvroSchema(topicSchema: TopicSchema) =
...
-        let scemaVersionBytes =
-            topicSchema.SchemaVersion
+        let schemaVersionBytes =
+            schemaVersion
             |> Option.map (fun sv -> sv.Bytes)
             |> Option.toObj
-        GenericRecord(scemaVersionBytes, fields)
+        GenericRecord(schemaVersionBytes, fields)
+        
+    override this.GetSpecificSchema schemaInfo schemaVersion =
```
Lanayx: Can you please make it a tuple (since ISchema is public and C# doesn't support curried parameters well)

Lanayx (approved):

Issue #188, opened 2021-09-14
Release for enum schema type supprot


RobertIndie:
Since https://github.com/eaba/AvroSchemaGenerator/pull/19 has been merged and AvroSchemaGenerator 2.2.0 has been released. Could we cut a new release for this client? @Lanayx

Lanayx:
Hi, @RobertIndie ! There is no need to cut another release, dependency can be updated separately.

RobertIndie:
Get it. Thanks!

PR #190, opened 2021-09-15
Update AvroSchemaGenerator to 2.3

nodece:
Signed-off-by: Zixuan Liu <nodeces@gmail.com>

### Summary

The `AvroSchemaGenerator@1.9` doesn't check the type.Namespace, when type.Namespace is null, the generated schema is incorrect, so I made a PR: https://github.com/eaba/AvroSchemaGenerator/pull/22 to fix this issue, this patch has been released in `AvroSchemaGenerator@2.3`.


Lanayx:
Hi, @nodece ! There is no need to update dependencies each time they change, you can just update the AvroSchemaGenerator nuget package in your project. Same as #188

PR #195, opened 2021-11-10
Add Wellknown data type support for ProtobufNativeSchema

### LINKED ISSUE #194: DateTime is not supported when using Protobuf native schema
We found a problem with F# client parsing protobuf native with `DateTime` type field.
In the F# client, the VirtualFile implementation only supports reading a single proto file, but not other dependent proto files.
Here is the current VirtualFile implementation:
```fsharp
type VirtualFile(content:string)=
    member private this.Content = content
    interface IFileSystem with
        member this.Exists _ = true
        member this.OpenText _ =
            new StringReader(this.Content) :> TextReader
```
As an example, if we use the following class to generate a protobuf definition.
```csharp
[ProtoContract]
class SimpleProtoRecordV1
{
    [ProtoMember(1, IsRequired = true), Required]
    public string Name { get; set; }
    [ProtoMember(2), DefaultValue(30)]
    public int Age { get; set; }
    [ProtoMember(3, IsRequired = false)]
    [CompatibilityLevel(CompatibilityLevel.Level300)]
    public DateTime Time { get; set; }
}
```
The `DateTime` type depends on [`/google/protobuf/timestamp.proto`](https://github.com/protobuf-net/protobuf-net/blob/14da8b12ed/src/protobuf-net.Reflection/google/protobuf/timestamp.proto), so the FileDescriptorSet will try to read `/google/protobuf/timestamp.proto` from the VirtualFile, but the wrong VirtualFile implementation causes it to return only the original SimpleProtoRecordV1 protobuf definition, making `/google/protobuf/timestamp.proto`'s dependency is still itself. 
This causes a stack overflow on the broker side when parsing this circular dependency. 

I have created a PR to fix this problem: https://github.com/fsharplang-ru/pulsar-client-dotnet/pull/195

RobertIndie:
This fixes #194 

Currently, ProtobufNativeSchema does not support handling data with builtin well-known type like DateTime. This PR will add Wellknown type(include `System.DateTime`, `System.TimeSpan`, `System.Guid`and `System.Decimal`) support for ProtobufNativeSchema. 

> Currently, this PR is only a preliminary change, the specific implementation solution still needs to be discussed. I am wondering if there is a better solution so that we don't need to introduce the content of proto files(such as `/google/protobuf/timestamp.proto`) in our project. Any ideas on this ? @Lanayx  Thanks!

Updated: This PR is ready for review.

### DIFF
--- a/src/Pulsar.Client/Schema/ProtobufNativeSchema.fs
+++ b/src/Pulsar.Client/Schema/ProtobufNativeSchema.fs
@@ -20,11 +20,34 @@ type   ProtobufNativeSchemaData(fileDescriptorSet:byte[],
     member  this.rootMessageTypeName =rootMessageTypeName     
     member  this.rootFileDescriptorName =rootFileDescriptorName
 type VirtualFile(content:string)=
+    let timestampProtoName = "/google/protobuf/timestamp.proto"
+    let timestampProtoContent = "
+syntax = \"proto3\";
+
+package google.protobuf;
+
+option csharp_namespace = \"Google.Protobuf.WellKnownTypes\";
+option cc_enable_arenas = true;
+option go_package = \"google.golang.org/protobuf/types/known/timestamppb\";
+option java_package = \"com.google.protobuf\";
+option java_outer_classname = \"TimestampProto\";
+option java_multiple_files = true;
+option objc_class_prefix = \"GPB\";
+
+message Timestamp {
+  int64 seconds = 1;
+  int32 nanos = 2;
+}
+"
     member private this.Content = content
     interface IFileSystem with
         member this.Exists _ = true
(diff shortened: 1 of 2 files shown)


Lanayx:
Hi, @RobertIndie ! I think we should be able to load proto files content from the protobuf-reflection library, as soon as they are already loaded as embedded resources
image
Can you try using this approach instead of hardcoding the content?

RobertIndie:
Great! I will try.

RobertIndie:
Hi @Lanayx . This PR is ready for review. Thanks!

Lanayx (approved):

Lanayx:
Thanks, @RobertIndie ! I'll do a little cleanup and publish a version soon

PR #198, opened 2021-11-24
Fix set the replicationClusters param

nodece:
There are should get the  `ReplicationClusters ` from the message body, otherwise always `null`.

### DIFF
--- a/src/Pulsar.Client/Internal/BatchMessageContainer.fs
+++ b/src/Pulsar.Client/Internal/BatchMessageContainer.fs
@@ -104,7 +104,6 @@ type internal MessageContainer<'T>(config: ProducerConfiguration) =
     member val MaxMessageSize = Commands.DEFAULT_MAX_MESSAGE_SIZE with get, set
     member val NumMessagesInBatch = 0 with get, set
     member val CurrentTxnId = None with get, set
-    member val ReplicationClusters = None with get, set
 
 open BatchHelpers
 
@@ -128,7 +127,7 @@ type internal DefaultBatchMessageContainer<'T>(prefix: string, config: ProducerC
             PartitionKey = batchItems.[0].Message.Key
             OrderingKey = batchItems.[0].Message.OrderingKey
             TxnId = this.CurrentTxnId
-            ReplicationClusters = this.ReplicationClusters
+            ReplicationClusters = batchItems.[0].Message.ReplicationClusters
         }
     override this.CreateOpSendMsgs () =
         raise <| NotSupportedException()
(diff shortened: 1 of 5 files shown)


Lanayx:
Hi, thank you for the PR, but the build is failing. Can you also remove[ this property](https://github.com/fsharplang-ru/pulsar-client-dotnet/blob/develop/src/Pulsar.Client/Internal/BatchMessageContainer.fs#L107) as it will be unused

src/Pulsar.Client/Internal/BatchMessageContainer.fs
```diff
@@ -128,7 +128,7 @@ type internal DefaultBatchMessageContainer<'T>(prefix: string, config: ProducerC
             PartitionKey = batchItems.[0].Message.Key
             OrderingKey = batchItems.[0].Message.OrderingKey
             TxnId = this.CurrentTxnId
-            ReplicationClusters = this.ReplicationClusters
+            ReplicationClusters = batchItems.[0].ReplicationClusters
```
RobertIndie: This seems to work only for the replication clusters of the first message. If the first message is specified to be replicated to cluster A, then the second message is specified to be replicated to cluster B. The entire batch of messages does not appear to be replicated to cluster B, and this behavior should not be what we expect. Please correct me if I am wrong.

RobertIndie (commented):
Could you add some tests to verify the changes?

src/Pulsar.Client/Internal/BatchMessageContainer.fs
nodece: The `batchItems` should be a queue, and have many messages, we will remove the first element from this queue after calling `CreateOpSendMsg`, so it should be working.

Lanayx: @RobertIndie as far as I can see in the java code it behaves exactly the same - replication settings are just taken from the first message in a batch

RobertIndie: @Lanayx In the java client, the replication clusters of the whole batch message are take from all single messages in batch. [Here is the code](https://github.com/apache/pulsar/blob/877bf3a34e72336c05706ffd48826e4347606196/pulsar-common/src/main/java/org/apache/pulsar/common/protocol/Commands.java?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L1614-L1616) But in the java client, this behavior should also be incorrect.
We need to get the messages of the same replication clusters into the same batch. @nodece Could we also fix this issue in this PR? I can give a fix on the java client side.

nodece: I can fix this issue on next PR.

Lanayx: @RobertIndie that's right, the code that you sent is only called for the first message of the batch, so all other messages in the batch will inherit that settings. I agree with @nodece  that we can fix the behavior in the next PR when you are ready with Java code changes.

Lanayx (approved):

PR #203, opened 2022-02-19
Replace testAsync to testTask in unit and integration tests

Maxumka:
Hi, I replace testAsync to testTask. Also replaced Async.Sleep to Task.Delay and added types annotation for some let! bindings like consumer, producer and message, because task computation expression does not inference types and complier show that error "FS0072: Lookup on object of indeterminate type"

### DIFF
--- a/tests/IntegrationTests/Acks.fs
+++ b/tests/IntegrationTests/Acks.fs
@@ -7,6 +7,8 @@ open Expecto.Flip
 open System.Text
 open System.Threading.Tasks
 open System.Threading
+open FSharp.UMX
+open Pulsar.Client.Api
 open Pulsar.Client.Common
 open Serilog
 open System.Collections.Generic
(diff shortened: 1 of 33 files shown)


Lanayx:
Thank you for the PR! I'll review it tomorrow or on Monday. In the meantime you could try create an issue in fsharp repository to improve type inference for tasks

Lanayx:
Can you please change all ptestAsync to ptestTask, tests like Eternal loop, 2К of topics, Failover consumer with PriorityLevel are probably causing timeout

Lanayx:
It seems one test is failing now, can you please take a look?
изображение

Maxumka:
Test "Concurrent send and receive work fine" from integration test, Basic.fs has try-with and first case checks type inner exception, 239 line, but inner exp can be null, maybe just added null check for inner exception?

Lanayx:
Well, the idea of this test is to check whether AlreadyClosedException is thrown or not, the null check won't give us the answer. If you could run the test locally and see what is the actual type of ex, this would definetely give us the answer. My guess is that after changing async to task the exception is no longer wrapped in AggregateException, so rather than checking type of inner exception we can check the type of exception itselft.

Maxumka:
You are right, with testTask the exception is no longer wrapped in AggregateException, and its type AlreadyClosedException

Lanayx (approved):

PR #204, opened 2022-02-21
feat: add TableView support

nodece:
Signed-off-by: Zixuan Liu <nodeces@gmail.com>

### Summary

The TableView was introduced in Pulsar 2.10, we need this feature in pulsar-client-dotnet.

Proposal: https://github.com/apache/pulsar/issues/12356
The Java implementation: https://github.com/apache/pulsar/pull/12838

### DIFF
--- a/src/Pulsar.Client/Api/ITableView.fs
+++ b/src/Pulsar.Client/Api/ITableView.fs
@@ -1,0 +1,8 @@
+namespace Pulsar.Client.Api
+
+open System
+open System.Collections.Generic
+
+type ITableView<'T> =
+    inherit IAsyncDisposable
+    inherit IReadOnlyDictionary<String, 'T>
--- a/src/Pulsar.Client/Internal/TableViewImpl.fs
+++ b/src/Pulsar.Client/Internal/TableViewImpl.fs
@@ -13,7 +13,7 @@ type internal TableViewImpl<'T> private (reader: IReader<'T>) =
     
     member private this.HandleMessage(msg: Message<'T>) = 
         if not (String.IsNullOrEmpty(%msg.Key)) then
-            data.TryAdd(%msg.Key, msg.GetValue()) |> ignore
+            data.AddOrUpdate(%msg.Key, msg.GetValue(), (fun _ _ -> msg.GetValue())) |> ignore
 
     member private this.ReadTailMessages(reader: IReader<'T>) = 
         task {
--- a/src/Pulsar.Client/Internal/TableViewImpl.fs
+++ b/src/Pulsar.Client/Internal/TableViewImpl.fs
@@ -1,0 +1,74 @@
+namespace Pulsar.Client.Internal
+
+open System.Collections.Concurrent
+open System.Threading.Tasks
+open FSharp.UMX
+open Pulsar.Client.Common
+open Pulsar.Client.Api
+
+open System
+
+type internal TableViewImpl<'T> private (reader: IReader<'T>) =
+    let data = ConcurrentDictionary<String, 'T>()
+    
+    member private this.HandleMessage(msg: Message<'T>) = 
+        if not (String.IsNullOrEmpty(%msg.Key)) then
+            data.TryAdd(%msg.Key, msg.GetValue()) |> ignore
+
+    member private this.ReadTailMessages(reader: IReader<'T>) = 
+        task {
+            let! msg = reader.ReadNextAsync()
+            this.HandleMessage(msg)
+            this.ReadTailMessages(reader) |> ignore
+        }
+
+    member private this.ReadAllExistingMessages(reader: IReader<'T>) = 
+        task {
+            let! hasMessage = reader.HasMessageAvailableAsync()
+            if hasMessage then
+                let! msg = reader.ReadNextAsync()
+                this.HandleMessage(msg)
+                return! this.ReadAllExistingMessages(reader)
+            else
+                this.ReadTailMessages(reader) |> ignore
+                return 0;
+        }
+
+    static member internal Init(createReader: _ -> Task<IReader<'T>>) =
+        task {
+            let! reader = createReader()
+            let tableView = TableViewImpl(reader)
+            tableView.ReadAllExistingMessages(reader).Wait()
+            return tableView
+        }
+
+    interface ITableView<'T> with
+        member this.Count
+            with get () = data.Count
+
+         member this.Keys
+            with get () = data.Keys
+
+        member this.Values
+            with get () = data.Values
+
+        member this.Item
+           with get(key) = data[key]
+
+        member this.GetEnumerator() =
+            data.GetEnumerator()
+
+        member this.GetEnumerator(): Collections.IEnumerator =
+            data.GetEnumerator()
+
+        member this.ContainsKey(key) =
+            data.ContainsKey(key)
+
+        member this.TryGetValue(key, value) =
+            data.TryGetValue(key, &value)
+    
+    interface IAsyncDisposable with
+       member this.DisposeAsync() =
+           task {
+               reader.DisposeAsync() |> ignore
+           } |> ValueTask
--- a/tests/IntegrationTests/TableView.fs
+++ b/tests/IntegrationTests/TableView.fs
@@ -1,0 +1,57 @@
+module Pulsar.Client.IntegrationTests.TableView
+
+open System
+open System.Linq
+open Expecto
+open Expecto.Flip
+open Pulsar.Client.Api
+
+open Serilog
+open Pulsar.Client.IntegrationTests.Common
+
+[<Tests>]
+let tests =
+    testList "TableView1" [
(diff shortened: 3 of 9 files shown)


src/Pulsar.Client/Internal/TableViewImpl.fs
```diff
@@ -0,0 +1,78 @@
...
+
+open System.Threading.Tasks
+open FSharp.UMX
+open Pulsar.Client.Common
+open Pulsar.Client.Api
+
+open System
+
+type internal TableViewImpl<'T> private () =
+    let data = Map.empty<String, 'T>
```
Lanayx: Fsharp Map is an immutable structure, a new map is created on each change. Using Dictionary will be much more performant. Moreover, since data will be updated from multiple threads, we might want to use ConcurrentDictionary.

src/Pulsar.Client/Internal/TableViewImpl.fs
```diff
@@ -0,0 +1,78 @@
...
+    let apply1 (transform : unit -> IReader<'T> ) y = transform y
+
+    let readers = Map.empty<String, IReader<'T>>
+    
+    member private this.HandleMessage(msg: Message<'T>) = 
+        if not (String.IsNullOrEmpty(%msg.Key))
```
Lanayx: Fsharp style guide is 
```
if ... then
    ...
else
    ...
```

src/Pulsar.Client/Internal/TableViewImpl.fs
```diff
@@ -0,0 +1,78 @@
...
+
+    member private this.ReadAllExistingMessages(reader: IReader<'T>) = 
+        task {
+            let! hasMessage = reader.HasMessageAvailableAsync()
+            if hasMessage then
+                let! msg = reader.ReadNextAsync()
+                this.HandleMessage(msg)
+                return! this.ReadAllExistingMessages(reader)
+            else
+                Task.Run(fun ()->
```
Lanayx: Task.Run is not needed here

src/Pulsar.Client/Api/ITableView.fs
```diff
@@ -0,0 +1,28 @@
...
+    abstract member Size: unit -> int
+    
+    // Whether the TableView has the key-value mappings.
+    abstract member IsEmpty: unit -> bool
+    
+    // Whether the TableView contains this given key.
+    abstract member ContainsKey: string -> bool
+    
+    // Gets key-value mapping by the given key.
+    abstract member Get: string -> 'T option
```
Lanayx: Option type is not C#-frendly, it's better to user ValueOption

src/Pulsar.Client/Internal/TableViewImpl.fs
```diff
@@ -0,0 +1,78 @@
...
+open Pulsar.Client.Common
+open Pulsar.Client.Api
+
+open System
+
+type internal TableViewImpl<'T> private () =
+    let data = Map.empty<String, 'T>
+    let apply1 (transform : unit -> IReader<'T> ) y = transform y
+
+    let readers = Map.empty<String, IReader<'T>>
```
Lanayx: Should also be ConcurrentDictionary

Lanayx (changes requested):
The warnings require to specify return type directly.
Also please update pulsar version if needed [here](https://github.com/fsharplang-ru/pulsar-client-dotnet/blob/develop/tests/compose/standalone/docker-compose.yml#L9)

nodece:
Thanks @Lanayx for your review, but the `.NET Core Ubuntu / build` CI failed, could you give me some help?

Lanayx:
Here is the error, looks like you need to update pulsar version as I mentioned in previous reply
изображение

src/Pulsar.Client/Api/ITableView.fs
```diff
@@ -0,0 +1,28 @@
+namespace Pulsar.Client.Api
+
+open System
+open System.Collections.Generic
+
+type ITableView<'T> =
```
Lanayx, line 6: I've just got an idea - instead of making all those custom abstract members you can just inherit from IReadOnlyDictionary interface and implement it's members

nodece: The `Keys` and `Values` should be methods, but in `IReadOnlyDictionary`, it is a property, I don't think use the `IReadOnlyDictionary` can implement this.

Lanayx: Right, I think they should be properties rather than methods :) The underlying ConcurrentDictionary has them as properties as well. In .NET properties are widely used for such things instead of methods

nodece: Thanks,  please review.

tests/IntegrationTests/TableView.fs
```diff
@@ -0,0 +1,57 @@
...
+open Pulsar.Client.Api
+
+open Serilog
+open Pulsar.Client.IntegrationTests.Common
+open FSharp.Control
+
+[<Tests>]
+let tests =
+    testList "TableView" [
+        testAsync "TableView with non-partition-topic works fine" {
```
Lanayx: Can you please change to testTask and remove `|>Async.AwaitTask`  in the test body

tests/IntegrationTests/TableView.fs
```diff
@@ -0,0 +1,57 @@
...
+            let producerName = "tableViewProducer"
+            let topicName = "public/default/" + Guid.NewGuid().ToString("N")
+
+            let! producer =
+                client.NewProducer()
+                    .Topic(topicName)
+                    .ProducerName(producerName)
+                    .CreateAsync()  |> Async.AwaitTask
+
+            do! produceMessagesWithProps producer numberOfMessages producerName  |> Async.AwaitTask
```
Lanayx: Why does this work? Usually producing to topic without consumer results in just discarding messages. 
Anyway, another idea is that it will be more useful to produce messages with specific key+value and check them below. For example key1+value1, key1+value2, key2+value3. And then check both keys below

tests/IntegrationTests/TableView.fs
```diff
@@ -0,0 +1,57 @@
...
+            Expect.equal "" numberOfMessages tableView.Count
+
+            let keysCount = Enumerable.Count(tableView.Keys)
+            Expect.equal "" numberOfMessages keysCount
+
+            let valuesCount = Enumerable.Count(tableView.Values)
+            Expect.equal "" numberOfMessages valuesCount
+
+            let contained = tableView.ContainsKey(tableView.Keys.First())
+            Expect.equal "" true contained
```
Lanayx, line 49: Rather than checking first key it makes more sense to check for specific key

tests/IntegrationTests/TableView.fs
```diff
@@ -0,0 +1,57 @@
...
+            
+            let enumerator = tableView.GetEnumerator()
+            Expect.isNotNull "" enumerator
+            
+            let firstValue = tableView[tableView.Keys.First()]
+            Expect.isNotNull "" firstValue
```
Lanayx: Rather than checking first key it makes more sense to check for specific key and it's value

Lanayx (changes requested):

tests/IntegrationTests/TableView.fs
nodece: We only push the message with the key to `TableView`, the message without the key will be discarded.

Lanayx: Yes, I meant that messages are produced to the topic before consumer (reader -> table view) is created. In all other tests messages are only produced after consumer creation.

tests/IntegrationTests/TableView.fs
nodece: When I use the `testTask`, throw the following log:
```
Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved.
```
```
No overloads match for method 'Bind'.

Known types of arguments: Threading.Tasks.Task<ITableView<byte[]>> * (('a list -> 'b) -> Tasks.CopiedDoNotReference.TaskBuilder.Step<unit>) when 'b: null
```

The `tableView.Count` doesn't work, could you have any ideas?

nodece: The F# compile order is disgusting :(

Lanayx: Sorry, I'll be able to only try it in the evening today. In the meantime you can look at other tests - we transitioned to testTask everywhere just several days ago

nodece: The issue has been fixed, thanks.

Lanayx:
@nodece right now the test fails, since the old value is stored in the dictionary rather than a new value, is it expected?
изображение

nodece:
@Lanayx Good catch! This test has been fixed, use `AddOrUpdate()` instead `TryAdd()`.

Lanayx (approved):

PR #205, opened 2022-03-02
Fix seek chunked messages

RobertIndie:
## Motivation

This is the implementation of https://github.com/apache/pulsar/issues/12402.

Currently, when we send chunked messages, the producer returns the message-id of the last chunk. This can cause some problems. For example, when we use this message-id to seek, it will cause the consumer to consume from the position of the last chunk, and the consumer will mistakenly think that the previous chunks are lost and choose to skip the current message. If we use the inclusive seek, the consumer may skip the first message, which brings the wrong behavior.

Here is the simple code(in java) used to demonstrate the problem.

```java
var msgId = producer.send(...); // eg. return 0:1:-1

var otherMsg = producer.send(...); // return 0:2:-1

consumer.seek(msgId); // inclusive seek

var receiveMsgId = consumer.receive().getMessageId(); // it may skip the
first message and return like 0:2:-1

Assert.assertEquals(msgId, receiveMsgId); // fail
```

For more context, please see [PIP-107](https://github.com/apache/pulsar/issues/12402)

And I find that f# client has already stored all chunk message ids in MessageIds.chunkMessageIds. We can use this field to implement the ChunkMessageId feature like in java. 

There is still work left in this PR to serialize the ChunkMessageId. To be consistent with the behavior of the Java client, when we serialize and deserialize messageIDs or compare messageId, the comparison for chunkMessageIds only needs to compare the message id of the first chunk if the message is a chunked message. Like below:

```fsharp
match m.ChunkMessageIds, this.ChunkMessageIds with
| Some mchunkMessageIds, Some thisChunkMessageIds when mchunkMessageIds.Length > 0 && thisChunkMessageIds.Length > 0 ->
                        mchunkMessageIds.[0] = thisChunkMessageIds.[0] // We need to check the first chunk message id if the message is a chunkd message
| _, _ -> true
```

We need to update the pulsar proto file before proceeding with the rest of the work. What is the correct way to generate the code for the proto? I found that the code I generated using `protoc` is very different from the existing generated code. Are the parameters not set correctly?

Update: The serialization for the chunk message id is added. This PR is ready for review.

## Modification

* Fix consumer inclusive seek for chunked message
* Add compare for the first chunk message id in MessageId.

### DIFF
--- a/tests/IntegrationTests/Chunks.fs
+++ b/tests/IntegrationTests/Chunks.fs
@@ -143,4 +143,49 @@ let tests =
              
             Log.Debug("Ended Two parallel chunks-message delivered successfully with short queue")
         }
+
+        testTask "Seek chunk messages and receive correctly" {
+            Log.Debug("Started Seek chunk messages and receive correctly")
+            let client = getClient()
+            let topicName = "public/default/topic-" + Guid.NewGuid().ToString("N")
+            let name = "seekChunkMessages"
+
+            let! (producer : IProducer<byte[]>) =
+                client.NewProducer()
+                    .Topic(topicName)
+                    .ProducerName(name)
+                    .EnableBatching(false)
+                    .EnableChunking(true)
+                    .CompressionType(CompressionType.Snappy)
+                    .CreateAsync() 
+
+            let! (consumer :  IConsumer<byte[]>) =
+                client.NewConsumer()
+                    .Topic(topicName)
+                    .ConsumerName(name)
+                    .StartMessageIdInclusive()
+                    .SubscriptionName("test-subscription")
+                    .SubscribeAsync() 
+
+            let payload = Array.zeroCreate 10_000_000
+            Random().NextBytes(payload)
+            
+            let msgIds = [ for i in 0 .. 9 -> 
+                            producer.NewMessage(payload)
+                            |> producer.SendAsync
+                            |> Async.AwaitTask
+                            |> Async.RunSynchronously]
+
+            for i in 0 .. 9 do
+                do! consumer.ReceiveAsync()
+
+            do! consumer.SeekAsync(msgIds.[1])
+            for i in 1 .. 9 do
+                let! (msgAfterSeek : Message<byte[]>) = consumer.ReceiveAsync()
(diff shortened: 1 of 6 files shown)


Lanayx:
The code is generated using this site https://protogen.marcgravell.com/ , you'll also need to update generated modifiers from public to internal.

tests/IntegrationTests/Chunks.fs
```diff
@@ -143,4 +143,49 @@ let tests =
...
+                    .SubscribeAsync() 
+
+            let payload = Array.zeroCreate 10_000_000
+            Random().NextBytes(payload)
+            
+            let msgIds = [ for i in 0 .. 9 -> 
+                            producer.NewMessage(payload)
+                            |> producer.SendAsync
+                            |> Async.AwaitTask
+                            |> Async.RunSynchronously]
```
Lanayx: Can you please rewrite it similar to [this](https://github.com/fsprojects/pulsar-client-dotnet/blob/develop/tests/IntegrationTests/Chunks.fs#L105-L108)

tests/IntegrationTests/Chunks.fs
```diff
@@ -143,4 +143,49 @@ let tests =
...
+            for i in 0 .. 9 do
+                do! consumer.ReceiveAsync()
+
+            do! consumer.SeekAsync(msgIds.[1])
+            for i in 1 .. 9 do
+                let! (msgAfterSeek : Message<byte[]>) = consumer.ReceiveAsync()
+                Expect.equal "" msgIds.[i] msgAfterSeek.MessageId
+        
+            do! consumer.UnsubscribeAsync() 
+            do! Task.Delay 100
```
Lanayx: This delay is no longer needed (since unsubscribe waits for pending acks since 2.9.4)

tests/IntegrationTests/Chunks.fs
RobertIndie: I didn't find a good way to rewrite it. I got some compile errors when I tried it, Could you give me some guidance?

Lanayx (approved):

PR #233, opened 2023-09-29
Add a missing mapping between Pulsar AuthChallenge command and its ha…

LeoSht:
Hello,
I noticed that the Pulsar AuthChallenge command is not mapped to its handler, possibly just a typo?

Pulsar sends AuthChallenge command 1 minute before the Auth token expiration. Response should contain the new token.

Once mapped, the existing handler seems to work fine. Otherwise, the command is ignored, and Pulsar drops the connection when token expires. Tested manually with Token authentication.

Regards,
Leo.


Lanayx (approved):
Thanks, great observation!

Issue #239, opened 2023-11-16
Support isConnected for producers, consumers and readers


Lanayx:
Java PR [link](https://github.com/apache/pulsar/pull/1780)

baumerik:
@Lanayx: Please check if it meet your expectation.

PR #240, opened 2023-11-17
expose the connection state to consumer and producer interface #239

baumerik:
Hi @Lanayx,

here is my attempt for exposing the `IsConnected`. 
For the `MultiTopicsConsumerImpl` and the `PartitionedProducerImpl` I've assumed that it will be return `true` just when all Topics are connected and ready.
I've tested it localy it with the `MultiTopicsConsumerImpl` and the property switched es expected.
please give me feedback if it fits to your expectation.

Thanks
Erik

### DIFF
--- a/src/Pulsar.Client/Internal/ConsumerImpl.fs
+++ b/src/Pulsar.Client/Internal/ConsumerImpl.fs
@@ -1664,6 +1664,11 @@ type internal ConsumerImpl<'T> (consumerConfig: ConsumerConfiguration<'T>, clien
 
         member this.LastDisconnectedTimestamp =
             connectionHandler.LastDisconnectedTimestamp
+            
+        member this.IsConnected =
+            match connectionHandler.ConnectionState with
+                | Ready _ -> true
+                | _ -> false
 
 
     interface IAsyncDisposable with
--- a/src/Pulsar.Client/Internal/ConsumerImpl.fs
+++ b/src/Pulsar.Client/Internal/ConsumerImpl.fs
@@ -1667,8 +1667,8 @@ type internal ConsumerImpl<'T> (consumerConfig: ConsumerConfiguration<'T>, clien
             
         member this.IsConnected =
             match connectionHandler.ConnectionState with
-                | Ready _ -> true
-                | _ -> false
+            | Ready _ -> true
+            | _ -> false
 
 
     interface IAsyncDisposable with
--- a/src/Pulsar.Client/Internal/MultiTopicsConsumerImpl.fs
+++ b/src/Pulsar.Client/Internal/MultiTopicsConsumerImpl.fs
@@ -62,6 +62,7 @@ type internal MultiTopicConsumerMessage<'T> =
     | CancelWaiter of Waiter<'T>
     | CancelBatchWaiter of BatchWaiter<'T>
     | LastDisconnectedTimestamp of TaskCompletionSource<TimeStamp>
+    | IsConnected of TaskCompletionSource<bool>
     | HasMessageAvailable of TaskCompletionSource<bool>
 
 type internal TopicAndConsumer<'T> =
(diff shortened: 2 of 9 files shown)


Lanayx:
Thanks for the PR, it looks good! Can you please also add it to the Reader interface and implementation, just like in Java code

src/Pulsar.Client/Internal/PartitionedProducerImpl.fs
```diff
@@ -197,6 +198,15 @@ type internal PartitionedProducerImpl<'T> private (producerConfig: ProducerConfi
                 |> Seq.max
                 |> channel.SetResult
 
+            | IsConnected channel ->
+
+                Log.Logger.LogDebug("{0} IsConnected", prefix)
+                producers
+                |> Seq.map (fun producer -> producer.IsConnected)
+                |> Seq.contains(false)
+                |> not
```
Lanayx: I think you can replace it with
`Seq.forall _.IsConnected`

src/Pulsar.Client/Internal/MultiTopicsConsumerImpl.fs
```diff
@@ -806,6 +807,15 @@ type internal MultiTopicsConsumerImpl<'T> (consumerConfig: ConsumerConfiguration
...
                 |> Seq.max
                 |> channel.SetResult
+            
+            | IsConnected channel ->
+
+                Log.Logger.LogDebug("{0} IsConnected", prefix)
+                consumers
+                |> Seq.map (fun (KeyValue(_, (consumer, _))) -> consumer.IsConnected)
+                |> Seq.contains(false)
+                |> not
```
Lanayx: I think you can replace it with
`Seq.forall (fun (KeyValue(_, (consumer, _))) -> consumer.IsConnected)`

src/Pulsar.Client/Internal/ConsumerImpl.fs
```diff
@@ -1664,6 +1664,11 @@ type internal ConsumerImpl<'T> (consumerConfig: ConsumerConfiguration<'T>, clien
 
         member this.LastDisconnectedTimestamp =
             connectionHandler.LastDisconnectedTimestamp
+            
+        member this.IsConnected =
+            match connectionHandler.ConnectionState with
+                | Ready _ -> true
+                | _ -> false
```
Lanayx: Please ident `|` back to be right under `match`

src/Pulsar.Client/Internal/ProducerImpl.fs
```diff
@@ -905,6 +905,10 @@ type internal ProducerImpl<'T> private (producerConfig: ProducerConfiguration, c
 
         member this.LastDisconnectedTimestamp =
             connectionHandler.LastDisconnectedTimestamp
+        member this.IsConnected =
+            match connectionHandler.ConnectionState with
+                | Ready _ -> true
+                | _ -> false
```
Lanayx: Please ident `|` back to be right under `match`

Lanayx (changes requested):

src/Pulsar.Client/Internal/ProducerImpl.fs
baumerik: done

src/Pulsar.Client/Internal/ConsumerImpl.fs
baumerik: done

src/Pulsar.Client/Internal/MultiTopicsConsumerImpl.fs
baumerik: looks much better 😃

src/Pulsar.Client/Internal/PartitionedProducerImpl.fs
baumerik: this syntax does not work but I changed it to `forall`

baumerik:
> Thanks for the PR, it looks good! 
Thanks for my first Feedback to F# 😄 

> Can you please also add it to the Reader interface and implementation, just like in Java code
sorry, this i haven't see in the first time. It is now added.

Lanayx:
Merged! Do you want to do this for 3.0 branch as well? 2.x will not receive performance updates and refactoring in future

Lanayx:
https://www.nuget.org/packages/Pulsar.Client/2.15.0 - this was published

PR #254, opened 2024-03-06
Ensure all connections are closed regardless of exception

glebvsmir:
Hello, 
With some recent changes made that support various types of exceptions, including cancellation, it turned out that some connections were not being closed leading to a leak of connections. 
The solution is to add a nested try...finally code block to ensure that connections are always closed for all possible exceptions.

### DIFF
--- a/src/Pulsar.Client/Internal/ClientCnx.fs
+++ b/src/Pulsar.Client/Internal/ClientCnx.fs
@@ -808,7 +808,7 @@ and internal ClientCnx (config: PulsarClientConfiguration,
             let mutable continueLooping = true
             let reader = connection.Input
 
-            try                
+            try
                 while continueLooping do
                     let! result = reader.ReadAsync(socketReadCancellationTokenSource.Token)
                     let buffer = result.Buffer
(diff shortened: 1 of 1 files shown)


Lanayx:
Thanks for commit! I think that the change can be simplified just by padding `post operationsMb ChannelInactive` 4 spaces left, so it will be called regardless of exception

glebvsmir:
Simplified as you recommended. This has been re-tested and appears to behave the same under normal execution.

glebvsmir:
> Thanks for commit! I think that the change can be simplified just by padding `post operationsMb ChannelInactive` 4 spaces left, so it will be called regardless of exception

Simplified as you recommended. This has been re-tested and appears to behave the same under normal execution.

Lanayx:
I meant only 4 spaces left :) But you can leave it with 8 spaces left, in that case can you please also remove `post operationsMb ChannelInactive` from line 819 since it will be called twice

glebvsmir:
> I meant only 4 spaces left :) But you can leave it with 8 spaces left, in that case can you please also remove `post operationsMb ChannelInactive` from line 819 since it will be called twice

Good observation. Updated per your original suggestion to keep connection closing within the exception handling block. I prefer to leave the "else" behavior as is.

Lanayx:
Fix released in https://www.nuget.org/packages/Pulsar.Client/2.15.2

glebvsmir:
Thank you

Issue #257, opened 2024-03-27
Possible bug decompressing Zstd with no size in the compressed payload.


Nikolajls:
Hi ! 

Been a user of the library for some time now as well as other teams inside the company I work at use it as well.

We've discovered something strange where a topic can be read through by other Pulsar Clients, (Python, Java) but using this library with version `3.3` it fails

I've debugged through the codebase of the F# Client library connected to the topic and the issue we face is that the messages is compressed with Zstd where we get this exception.:
"Decompressed content size is not specified"

With the current codebase for `3.3` it looks like this:
The call to unwrap in ZstdNet is done using the default true value for `bufferSizePrecheck ` so that library try to check the decompressed content size as part of the payload.

![2024-03-27 15 24 39](https://github.com/fsprojects/pulsar-client-dotnet/assets/7819014/8e8dfca8-61d2-4fa1-b1aa-569e19c22449)

[Zstd library Unwrap method](https://github.com/skbkontur/ZstdNet/blob/7a4de80feb49b9810958c7def59d6642f1b1ba9f/ZstdNet/Decompressor.cs#L91)

However if I modify the F# code to use false and prevent the check from happening it works and we can read the message:
![2024-03-27 15 26 19](https://github.com/fsprojects/pulsar-client-dotnet/assets/7819014/664e045a-b0bf-4168-a77a-109bdfee98e8)

According to what I've read on [Zstd here](https://facebook.github.io/zstd/zstd_manual.html)
![2024-03-27 14 50 33](https://github.com/fsprojects/pulsar-client-dotnet/assets/7819014/1b879c82-8268-4f8d-9801-d2a5d86f1aa6)
It seems like the decompressed size is not required in the payload, so by always defaulting to true it fails to work on Zstd compressed payloads where the size is not present.

Would it be possible to add a way to configure if we want to always do this check so, it can be disabled by the user if they deem it neccesary?

I've no experience with F#, actually only ever read it by reading and learning this library - i've tried to look if I could make the changes myself, but it is not something that I'm entirely sure is feasible by me.

Lanayx:
[...]

PR #258, opened 2024-03-28
Decompressing with Zstd set bufferSizePrecheck flag to false

Nikolajls:
Closes issue https://github.com/fsprojects/pulsar-client-dotnet/issues/257 whichs contains more information.

By calling the ZstdNet unwrap method with false for `bufferSizePrecheck` messages with Zstd compressed data that does not contain the decompressed size in the payload will still work.

A test case has been added that validates it if the removal of the flag is done at some point or switch to excplit true.

We are using version `3.3.0` and would really love if it is feasible to get it bumped in patch for that one :)

image

### DIFF
--- a/tests/UnitTests/Internal/CompressionCodecTests.fs
+++ b/tests/UnitTests/Internal/CompressionCodecTests.fs
@@ -24,6 +24,7 @@ let tests =
     let helloLZ4 = System.Convert.FromBase64String "UEhlbGxv"
     let helloSnappy = System.Convert.FromBase64String "BRBIZWxsbw=="
     let helloZStd = System.Convert.FromBase64String "KLUv/SAFKQAASGVsbG8="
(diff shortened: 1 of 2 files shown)


tests/UnitTests/Internal/CompressionCodecTests.fs
```diff
@@ -79,4 +80,21 @@ let tests =
         test "Codec should make ZStd decoding" {
             helloZStd |> testDecode CompressionType.ZStd
         }
+
+        test "Zstd should decode content with non specififed size in the compressed payload" {
+          //The compressed string  above has been compressed with zstd using https://www.npmjs.com/package/zstd-codec
+          //It seems this does not specify the the decompressed size inthe payload
```
balticore:
```suggestion
          //It seems this does not specify the the decompressed size in the payload
```

Lanayx (approved):

Issue #253, opened 2024-03-01
Async token supplier / non-file-based oauth2


aeons:
It would be great if we could either 
1. supply an async token supplier, or
2. give the oauth2 credentials as configuration instead of a file

Lanayx:
@aeons Authentication is an abstract class, so you can write your own implementation of it using configuration. As for async version (I suppose you are talking about making [GetAuthData method](https://github.com/fsprojects/pulsar-client-dotnet/blob/develop/src/Pulsar.Client/Api/Authentication.fs#L16) async) - it could be a good idea, but it is initially sync in java, so Pulsar.Client just followed it to simplify understanding and now it would be a breaking change

aeons:
I get that I can write my own implementation, but since there already is one included, it would be nice to just be able to do 
```csharp
.Authentication(AuthenticationFactoryOAuth2.ClientCredentials(issuerUrl, audience, clientId, clientSecret))
```

For the sync vs async, getting an oauth2 token will always be async, so having the method be sync will always introduce blocking, but I do get the reasoning to follow the reference implementation.

chriscameron-vertexinc:
I was disappointed to find I can't supply a base64 encoded URI in the form:
`data:application/json;base64,<base64-encoded value>`

It looks like the AuthenticationFactoryOAuth2.ClientCredentials is trying to append the URI to the current working directory to find a file, and fails.

Lanayx:
@chriscameron-vertexinc You are welcome to send a PR with an improvement. Right now only the [LocalPath of URI is used ](https://github.com/fsprojects/pulsar-client-dotnet/blob/develop/src/Pulsar.Client/Auth/Oauth2/AuthenticationOauth2.fs#L104), but I think we can extend it with base64 encoded value support.

chriscameron-vertexinc:
@Lanayx I've opened a completely untested PR here https://github.com/fsprojects/pulsar-client-dotnet/pull/263

I'll try it out tomorrow, but in the meantime I'd appreciate any feedback. I haven't written any F# before!

Lanayx:
@chriscameron-vertexinc Thanks, good job for the first time F#! I've added some style recommendations and will be ready to merge once you confirm it works

Lanayx:
Published https://www.nuget.org/packages/Pulsar.Client/3.5.0 with the improvement

Lanayx:
Closing as completed

PR #263, opened 2024-05-15
feat: Add data:application/json;base64 support to AuthenticationOauth2

chriscameron-vertexinc:
Added support for using OAuth2 with a base64 encoded string instead of a file path.

Workaround for https://github.com/fsprojects/pulsar-client-dotnet/issues/253


src/Pulsar.Client/Auth/Oauth2/AuthenticationOauth2.fs
```diff
@@ -81,6 +80,32 @@ type internal AuthenticationOauth2(issuerUrl: Uri, audience: string, privateKey:
             return temp
         }
 
+    let base64DeserializeCreds data =
+        let bytes = Convert.FromBase64String(data)
+        let decodedData = System.Text.Encoding.UTF8.GetString(bytes)
+        JsonSerializer.Deserialize<Credentials>(decodedData)
+
+    let decodeAndDeserializeCreds (privateKey: Uri) =
+        let split = privateKey.LocalPath.Split(',', 2)
```
Lanayx: I would write it in this way:
 ```fsharp
 match privateKey.LocalPath.Split(',', 2) with
 | [| contentType, data |] when contentType = "application/json;base64" ->
     base64DeserializeCreds data
 | _ ->
     raise <| NotSupportedException $"Content type '{contentType}' is not supported."
 ```
 I think `.ToLowerInvariant()` could be skipped

src/Pulsar.Client/Auth/Oauth2/AuthenticationOauth2.fs
```diff
@@ -81,6 +80,32 @@ type internal AuthenticationOauth2(issuerUrl: Uri, audience: string, privateKey:
             return temp
         }
 
+    let base64DeserializeCreds data =
+        let bytes = Convert.FromBase64String(data)
```
Lanayx: this can be written as
```fsharp
data
|> Convert.FromBase64String
|> System.Text.Encoding.UTF8.GetString
|> JsonSerializer.Deserialize<Credentials>
```

src/Pulsar.Client/Auth/Oauth2/AuthenticationOauth2.fs
```diff
@@ -81,6 +80,32 @@ type internal AuthenticationOauth2(issuerUrl: Uri, audience: string, privateKey:
...
+        let contentType = split[0]
+        let data = split[1]
+        match contentType.ToLowerInvariant() with
+        | "application/json;base64" ->
+            base64DeserializeCreds(data)
+        | _ ->
+            raise <| NotSupportedException($"Content type '{contentType}' is not supported.")
+
+    let deserializeCreds (privateKey: Uri) =
+        backgroundTask {
```
Lanayx: I would write it as 
```fsharp
match privateKey.Scheme with
| "file" ->
    openAndDeserializeCreds privateKey.LocalPath
| "data" ->
    decodeAndDeserializeCreds privateKey |> Task.FromResult
| _ ->
    raise <| NotSupportedException $"Scheme '{privateKey.Scheme}' is not supported."
```
and would also skip `.ToLowerInvariant()`

src/Pulsar.Client/Auth/Oauth2/AuthenticationOauth2.fs
chriscameron-vertexinc: Done!

src/Pulsar.Client/Auth/Oauth2/AuthenticationOauth2.fs
chriscameron-vertexinc: Done, I had to add an extra case for when there are a different number of elements in the split

src/Pulsar.Client/Auth/Oauth2/AuthenticationOauth2.fs
chriscameron-vertexinc: I'm having a lot of trouble with this one. The mixing of sync and async is resulting in casting issues.

chriscameron-vertexinc:
@Lanayx I've been able to drop these changes into my application and verify that the base64 credentials decode is working.

Lanayx (approved):

DanBiscotti:
Hi I'm hoping to use this feature in a .NET Framework project using the 2.x branch, but it looks like it hasn't been merged into that branch. I was able to apply the changes to that branch, although I had to change one line:
`match credentialsUrl.LocalPath.Split(',', 2) with` -> ` match credentialsUrl.LocalPath.Split([| ',' |], 2) with`
I tested and it seems to work, is it ok to get this change on the 2.x branch as well?

Lanayx:
Yes, it is ok, you are welcome to send a PR

PR #282, opened 2024-11-21
Fix issue with message payload being disposed while resending

### LINKED ISSUE #281: The message payload may be disposed while resending
Version: 3.6.0

When the producer resending the message, it would throw the following error:
```
[16:29:52.717 WRN] clientCnx(7, LogicalAddres Unspecified/apps-broker-0-fccfb098-3242-40e4-8d0c-2c15888ed404.usce1-whale.test.g.sn2.dev:6651) Socket was disconnected exceptionally on writing Send
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'The stream with Id ef744d42-7ea5-4652-9175-eddbd3bd3304 and Tag  is disposed.'.
   at Microsoft.IO.RecyclableMemoryStream.ThrowDisposedException() in /_/src/RecyclableMemoryStream.cs:line 1352
   at Microsoft.IO.RecyclableMemoryStream.Seek(Int64 offset, SeekOrigin loc) in /_/src/RecyclableMemoryStream.cs:line 1164
   at Pulsar.Client.Common.Commands.serializePayloadCommand@56.Invoke(PipeWriter output) in /Users/rbt/code/pulsar-client-dotnet/src/Pulsar.Client/Common/Commands.fs:line 84
   at <StartupCode$Pulsar-Client>.$ClientCnx.clo@302-18.MoveNext() in /Users/rbt/code/pulsar-client-dotnet/src/Pulsar.Client/Internal/ClientCnx.fs:line 304
```

The root cause is that the payload would be disposed regardless of whether the send is successful.
https://github.com/fsprojects/pulsar-client-dotnet/blob/e89bad8c00006e232f59f26b6b65a75c7c0d85a6/src/Pulsar.Client/Common/Commands.fs#L107
And if there is some issue when the message is inflight, like a network issue after sending the message. The producer will resend the message here:
https://github.com/fsprojects/pulsar-client-dotnet/blob/e89bad8c00006e232f59f26b6b65a75c7c0d85a6/src/Pulsar.Client/Internal/ProducerImpl.fs#L216

But now the message payload has already been disposed, so the clientCnx will fail to process the payload MemoryStream and throw an exception here:
https://github.com/fsprojects/pulsar-client-dotnet/blob/8ed324666074b7f9726a38338058b0632d3b635d/src/Pulsar.Client/Common/Commands.fs#L82

## Proposed solution

To address this issue, we should only dispose of the payload after receiving the SendReceipt and removing the message from the pendingMessages.

Maybe we can add the payload to the `PendingMessage` type like this:
```fsharp
type internal PendingMessage<'T> =
    {
        CreatedAt: TimeStamp
        SequenceId: SequenceId
        HighestSequenceId: SequenceId
        SendTask: SendTask // rename the original Payload field to `SendTask`
        Payload: MemoryStream // new field for the payload
        Callback : PendingCallback<'T>
    }
```

Avoid despoing the payload in the SendTask. And only depose the payload after pulling the pendingMessage from the queue:
https://github.com/fsprojects/pulsar-client-dotnet/blob/8ed324666074b7f9726a38338058b0632d3b635d/src/Pulsar.Client/Internal/ProducerImpl.fs#L600-L605

Could you help take a look? @Lanayx Thanks!

RobertIndie:
Fixes #281

## Motivation
When the producer resending the message, it would throw the following error:
```
[16:29:52.717 WRN] clientCnx(7, LogicalAddres Unspecified/apps-broker-0-fccfb098-3242-40e4-8d0c-2c15888ed404.usce1-whale.test.g.sn2.dev:6651) Socket was disconnected exceptionally on writing Send
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'The stream with Id ef744d42-7ea5-4652-9175-eddbd3bd3304 and Tag  is disposed.'.
   at Microsoft.IO.RecyclableMemoryStream.ThrowDisposedException() in /_/src/RecyclableMemoryStream.cs:line 1352
   at Microsoft.IO.RecyclableMemoryStream.Seek(Int64 offset, SeekOrigin loc) in /_/src/RecyclableMemoryStream.cs:line 1164
   at Pulsar.Client.Common.Commands.serializePayloadCommand@56.Invoke(PipeWriter output) in /Users/rbt/code/pulsar-client-dotnet/src/Pulsar.Client/Common/Commands.fs:line 84
   at <StartupCode$Pulsar-Client>.$ClientCnx.clo@302-18.MoveNext() in /Users/rbt/code/pulsar-client-dotnet/src/Pulsar.Client/Internal/ClientCnx.fs:line 304
```

The root cause is that the payload would be disposed regardless of whether the send is successful. And if there is some issue when the message is inflight, like a network issue after sending the message. The producer will resend the message. But now the message payload has already been disposed, so the clientCnx will fail to process the payload MemoryStream and throw an exception.

## Modification

- Add the payload to the PendingMessage
- Avoid despoing the payload in the SendTask. And only depose the payload after deque the pendingMessage from the queue

## Verification

I added a unit test for asserting the payload won't be deposed after `serializeDeserializePayloadCommand`. But it's hard to reproduce the reconnection in the unit test.

I have also verified that this PR could fix the issue: https://github.com/fsprojects/pulsar-client-dotnet/issues/281. I tested it by forcing the reconnecting and the producer could resend the messages successfully. Otherwise without this fix, it could sometimes get stuck resending messages due to the serialization error.

### DIFF
--- a/src/Pulsar.Client/Common/Commands.fs
+++ b/src/Pulsar.Client/Common/Commands.fs
@@ -104,7 +104,6 @@ let private serializePayloadCommand (command : BaseCommand) (metadata: MessageMe
             try
                 return! temp.CopyToAsync(output)
             finally
-                payload.Dispose()
                 temp.Dispose()
                 binaryWriter.Dispose()
         } :> Task
--- a/src/Pulsar.Client/Common/DTO.fs
+++ b/src/Pulsar.Client/Common/DTO.fs
@@ -437,7 +437,8 @@ type internal PendingMessage<'T> =
         CreatedAt: TimeStamp
         SequenceId: SequenceId
         HighestSequenceId: SequenceId
-        Payload: SendTask
+        SendTask: SendTask
+        Payload: MemoryStream
         Callback : PendingCallback<'T>
     }
 
--- a/src/Pulsar.Client/Internal/ProducerImpl.fs
+++ b/src/Pulsar.Client/Internal/ProducerImpl.fs
@@ -120,6 +120,7 @@ type internal ProducerImpl<'T> private (producerConfig: ProducerConfiguration, c
     let failPendingMessages (ex: exn) =
         while pendingMessages.Count > 0 do
             let msg = pendingMessages.Dequeue()
+            msg.Payload.Dispose()
             failPendingMessage msg ex
         while blockedRequests.Count > 0 do
             let struct(_, channel, _) = blockedRequests.Dequeue()
@@ -191,7 +192,7 @@ type internal ProducerImpl<'T> private (producerConfig: ProducerConfiguration, c
         pendingMessages.Enqueue(pendingMessage)
         match connectionHandler.ConnectionState with
         | Ready clientCnx ->
-            clientCnx.SendAndForget pendingMessage.Payload
+            clientCnx.SendAndForget pendingMessage.SendTask
         | _ ->
             Log.Logger.LogWarning("{0} not connected, skipping send", prefix)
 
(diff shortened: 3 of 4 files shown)


Lanayx (approved):

Lanayx:
Thanks! I've released [3.6.1](https://www.nuget.org/packages/Pulsar.Client/3.6.1) with this fix

Issue #295, opened 2025-02-26
Support lookUpService over http protocol


geniusjoe:
**Is your feature request related to a problem? Please describe.**

Currently Java and Golang pulsar sdk support httpLookUp service, so that we can use a pulsar admin url as its serviceUrl, i.e.
```
PulsarClientBuilder().ServiceUrl("http://localhost:8080").BuildAsync() # pulsar admin url
```
I think http lookup it's more flexible than `advertisedListener` configuration, as it can return different broker serviceUrls depending on various http request uri or http header. Since dotnet client does not support this feature yet, maybe we could realize it.

**Describe the solution you'd like**
Refering to Java and Golang lookup realization, we need to do the following steps:

1. Create an interface let's say `ILookupService`:
```
type ILookupService =
    let endPointResolver = EndPointResolver(config.ServiceAddresses)

    let resolveEndPoint() = endPointResolver.Resolve()

    abstract member GetPartitionsForTopic: TopicName -> Task<TopicName array>

    abstract member GetPartitionedTopicMetadata: CompleteTopicName -> Task<PartitionedTopicMetadata>

    abstract member GetBroker: CompleteTopicName -> Task<Broker>

    abstract member GetTopicsUnderNamespace NamespaceName * bool -> Task<string seq>

    abstract member GetSchema CompleteTopicName * SchemaVersion option -> Task<TopicSchema option>
```
2. Replace `BinaryLookupService` with `LookupService` in `PulsarClient.fs`, `ConsumerImpl.fs` etc.
3. Create a new http lookup service object to realize `LookupService`

**Additional context**
Java realization: 
https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/HttpLookupService.java 

Golang realiazation:
https://github.com/apache/pulsar-client-go/blob/020b55c1dce4475fe8781f4fb6f858de903844b8/pulsar/internal/lookup_service.go#L284

Lanayx:
Hi, httpLookupService wasn't implemented by design, as binaryLookup service was a newer and faster version that covers all the cases. It also simplified code in several places. I can mark your suggestion as an enhancement, but could you please provide your use case where you have to use http lookup and can't use binary lookup instead.

geniusjoe:
[...]

PR #297, opened 2025-03-02
feat: Add HttpLookupService to support http serviceUrl

geniusjoe:
support  https://github.com/fsprojects/pulsar-client-dotnet/issues/295

# Changes
- Add http protocol parse strategy in `ServiceUri`
- Add `ILookupService` as a lookup service interface
- Replace `BinaryLookupService` with `ILookupService` in internal classes declaration
- Add `HttpLookupService` integrated test cases

### DIFF
--- a/src/Pulsar.Client/Internal/HttpLookupService.fs
+++ b/src/Pulsar.Client/Internal/HttpLookupService.fs
@@ -1,0 +1,230 @@
+﻿namespace Pulsar.Client.Internal
+
+open System.Collections.Generic
+open System.IO
+open System.Net.Http
+open System.Text.Json.Serialization
+open Pulsar.Client.Api
+open Pulsar.Client.Common
+open System
+open System.Net
+open FSharp.UMX
+open System.Text.Json
+open Microsoft.Extensions.Logging
+open Pulsar.Client.Schema
+
+type internal HttpLookupService (config: PulsarClientConfiguration, _connectionPool: ConnectionPool) =
+
+    let httpClient = new HttpClient(new SocketsHttpHandler(
+        PooledConnectionLifetime = TimeSpan.FromMinutes(2),
+        AllowAutoRedirect = true
+    ))
+    let jsonOptions = JsonSerializerOptions(
+        PropertyNamingPolicy = JsonNamingPolicy.CamelCase
+     )
+    do jsonOptions.Converters.Add(JsonStringEnumConverter())
+
+    interface ILookupService with
+
+        member this.GetPartitionsForTopic (topicName: TopicName) =
+            backgroundTask {
+                let backoff =
+                    Backoff {
+                        Initial = TimeSpan.FromSeconds(5.0)
+                        MandatoryStop = config.OperationTimeout + config.OperationTimeout
+                        Max = TimeSpan.FromMinutes(1.0)
+                    }
+                let! metadata = this.GetPartitionedTopicMetadataInner(
+                    topicName.CompleteTopicName,
+                    backoff,
+                    int config.OperationTimeout.TotalMilliseconds
+                )
+                if metadata.Partitions > 0 then
+                    return Array.init metadata.Partitions topicName.GetPartition
+                else
+                    return [| topicName |]
+            }
+
+        member this.GetPartitionedTopicMetadata topicName =
+            backgroundTask {
+                let backoff =
+                    Backoff {
+                        Initial = TimeSpan.FromSeconds(5.0)
+                        MandatoryStop = config.OperationTimeout + config.OperationTimeout
+                        Max = TimeSpan.FromMinutes(1.0)
+                    }
+                let! result = this.GetPartitionedTopicMetadataInner(topicName, backoff, int config.OperationTimeout.TotalMilliseconds)
+                return result
+            }
+
+        member this.GetBroker(topicName : CompleteTopicName) =
+            this.GetBrokerInner(topicName)
+
+        member this.GetTopicsUnderNamespace (ns: NamespaceName, isPersistent: bool) =
+            backgroundTask {
+                let backoff =
+                    Backoff {
+                        Initial = TimeSpan.FromSeconds(5.0)
+                        MandatoryStop = config.OperationTimeout + config.OperationTimeout
+                        Max = TimeSpan.FromMinutes(1.0)
+                    }
+                let! result = this.GetTopicsUnderNamespaceInner(ns, backoff, int config.OperationTimeout.TotalMilliseconds, isPersistent)
+                return result
+            }
+
+        member this.GetSchema(topicName: CompleteTopicName, ?schemaVersion: SchemaVersion) =
+            backgroundTask {
+                let backoff =
+                    Backoff {
+                        Initial = TimeSpan.FromSeconds(5.0)
+                        MandatoryStop = config.OperationTimeout + config.OperationTimeout
+                        Max = TimeSpan.FromMinutes(1.0)
+                    }
+                let! result = this.GetSchemaInner(topicName, schemaVersion, backoff, int config.OperationTimeout.TotalMilliseconds)
+                return result
+            }
+
+    //  GET /admin/v2/{topic-domain}/{tenant}/{namespace}/{topic}/partitions?checkAllowAutoCreation=true
+    member private this.GetPartitionedTopicMetadataInner (topicName: CompleteTopicName, backoff: Backoff, remainingTimeMs) =
+         async {
+            try
+                let randomServiceUri = config.ServiceAddresses[RandomGenerator.Next(0, config.ServiceAddresses.Length)]
+                let topic: string = %topicName
+                let topicRestPath = topic.Replace("persistent://","persistent/").Replace("non-persistent://","non-persistent/")
+                let! response = randomServiceUri.AbsoluteUri + $"admin/v2/%s{topicRestPath}/partitions?checkAllowAutoCreation=true"
+                                |> httpClient.GetStreamAsync
+                                |> Async.AwaitTask
+                let brokerResponse = JsonSerializer.Deserialize<{| Partitions: int |}>(response, jsonOptions)
+                return { Partitions = brokerResponse.Partitions }
+
+            with Flatten ex ->
+                let nextDelay = Math.Min(backoff.Next(), remainingTimeMs)
+                if nextDelay <= 0 then
+                    reraize ex
+                Log.Logger.LogWarning(ex, "GetPartitionedTopicMetadata failed will retry in {0} ms", nextDelay)
+                do! Async.Sleep nextDelay
+                return! this.GetPartitionedTopicMetadataInner(topicName, backoff, remainingTimeMs - nextDelay)
+        }
+
+    //  GET /lookup/v2/topic/{topic-domain}/{tenant}/{namespace}/{topic}
+    member private this.GetBrokerInner(topicName: CompleteTopicName) =
+        backgroundTask {
+            let randomServiceUri = config.ServiceAddresses[RandomGenerator.Next(0, config.ServiceAddresses.Length)]
+            let topic: string = %topicName
+            let topicRestPath = topic.Replace("persistent://","persistent/").Replace("non-persistent://","non-persistent/")
+            let! response = httpClient.GetStreamAsync (randomServiceUri.AbsoluteUri + $"lookup/v2/topic/%s{topicRestPath}")
+            let brokerResponse = JsonSerializer.Deserialize<
+                                {|
+                                  brokerUrl : string
+                                  brokerUrlTls: string
+                                  httpUrl: string
+                                  httpUrlTls: string
+                                |}>(response, jsonOptions)
+            let uri = if config.UseTls
+                      then Uri(brokerResponse.brokerUrlTls)
+                      else Uri(brokerResponse.brokerUrl)
+            let resultEndpoint = DnsEndPoint(uri.Host, uri.Port)
+            return { LogicalAddress = LogicalAddress resultEndpoint; PhysicalAddress = PhysicalAddress resultEndpoint }
+        }
+
+    //  GET /admin/v2/namespaces/{tenant}/{namespace}/topics?mode=PERSISTENT
+    //  GET /admin/v2/namespaces/{tenant}/{namespace}/topics?mode=NON_PERSISTENT
+    member private this.GetTopicsUnderNamespaceInner (ns: NamespaceName, backoff: Backoff, remainingTimeMs: int, isPersistent: bool) =
+        async {
+            try
+                let randomServiceUri = config.ServiceAddresses[RandomGenerator.Next(0, config.ServiceAddresses.Length)]
+                let mode = match isPersistent with
+                            | true -> "PERSISTENT"
+                            | false -> "NON_PERSISTENT"
+                let! response = randomServiceUri.AbsoluteUri + $"admin/v2/namespaces/%s{ns.ToString()}/topics?mode=%s{mode}"
+                                |> httpClient.GetStreamAsync
+                                |> Async.AwaitTask
+                let brokerResponse = JsonSerializer.Deserialize<string seq>(response, jsonOptions)
+                return brokerResponse
+
+            with Flatten ex ->
+                let delay = Math.Min(backoff.Next(), remainingTimeMs)
+                if delay <= 0 then
+                    raise (TimeoutException "Could not getTopicsUnderNamespace within configured timeout.")
+                Log.Logger.LogWarning(ex, "GetTopicsUnderNamespace failed will retry in {0} ms", delay)
+                do! Async.Sleep delay
+                return! this.GetTopicsUnderNamespaceInner(ns, backoff, remainingTimeMs - delay, isPersistent)
+        }
+
+    //  GET /admin/v2/schemas/{tenant}/{namespace}/{topic}/schema
+    //  GET /admin/v2/schemas/{tenant}/{namespace}/{topic}/schema/{version}
+    member private this.GetSchemaInner(topicName: CompleteTopicName, schemaVersion: SchemaVersion option,
+                              backoff: Backoff, remainingTimeMs: int) =
+         async {
+            try
+                let randomServiceUri = config.ServiceAddresses[RandomGenerator.Next(0, config.ServiceAddresses.Length)]
+                let topic: string = %topicName
+                let topicRestPath = topic.Replace("persistent://","").Replace("non-persistent://","")
+                let path = match schemaVersion with
+                            | Some sv ->
+                                //  read 8 bytes from big endian
+                                let binaryWriter = new BinaryReader(new MemoryStream(sv.Bytes))
+                                let schemaVersionInt = binaryWriter.ReadInt64() |> int64FromBigEndian
+                                $"admin/v2/schemas/%s{topicRestPath}/schema/%d{schemaVersionInt}"
+                            | None ->
+                                $"admin/v2/schemas/%s{topicRestPath}/schema"
+                let! response = randomServiceUri.AbsoluteUri + path
+                                |> httpClient.GetStreamAsync
+                                |> Async.AwaitTask
+                let schemaResponse = JsonSerializer.Deserialize<
+                                    {|
+                                      Version : Int64
+                                      Type: SchemaType
+                                      Timestamp: Int64
+                                      Data: string
+                                      Properties: Dictionary<string, string>
+                                    |}>(response, jsonOptions)
+
+                let schemaData = match schemaResponse.Type with
+                                 | SchemaType.KEY_VALUE -> schemaResponse.Data |> this.getKeyValueSchemaBytes
+                                 | _ -> schemaResponse.Data |> System.Text.Encoding.UTF8.GetBytes
+                let schemaVersion: SchemaVersion = {
+                    Bytes = schemaResponse.Version |> BitConverter.GetBytes |> Array.rev
+                }
+                let schemaInfo: SchemaInfo = {
+                    Name = topicRestPath
+                    Type = schemaResponse.Type
+                    Properties = schemaResponse.Properties
+                    Schema = schemaData
+                }
+                let topicSchema: TopicSchema = {
+                    SchemaVersion = Some schemaVersion
+                    SchemaInfo = schemaInfo
+                }
+                return Some topicSchema
+
+            with
+                | Flatten ex ->
+                    match ex with
+                    //  When there is no topic related schema, pulsar http rest api will return a 404 exception
+                    //  In this case the request it's success, and we can return a None schema
+                    | :? HttpRequestException as ex when ex.StatusCode = Nullable HttpStatusCode.NotFound ->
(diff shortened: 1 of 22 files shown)


src/Pulsar.Client/Common/DTO.fs
```diff
@@ -142,19 +159,19 @@ type internal SendReceipt =
...
             $"LogicalAddres {l.ToString()}"
 
-type internal PhysicalAddress =
+type PhysicalAddress =
     PhysicalAddress of DnsEndPoint
         override this.ToString() =
             let (PhysicalAddress l) = this
             $"PhysicalAddress {l.ToString()}"
 
-type internal Broker =
```
Lanayx: Why have you removed internal modifiers in this file?

src/Pulsar.Client/Common/DTO.fs
```diff
@@ -54,6 +54,14 @@ type internal LookupTopicResult =
         Authoritative: bool
     }
 
+type HttpLookupTopicResult =
```
Lanayx: Should be internal

src/Pulsar.Client/Common/DTO.fs
```diff
@@ -94,7 +102,16 @@ type SchemaInfo =
         Properties: IReadOnlyDictionary<string, string>
     }
 
-type internal TopicSchema =
+type HttpSchemaInfo =
```
Lanayx: Should be internal

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -0,0 +1,203 @@
...
+open System
+open System.Net
+open FSharp.UMX
+open System.Text.Json
+open Microsoft.Extensions.Logging
+open Pulsar.Client.Schema
+
+type internal HttpLookupService (config: PulsarClientConfiguration, _connectionPool: ConnectionPool) =
+
+    let clientCnx = new HttpClient(new SocketsHttpHandler(
```
Lanayx: I'd call local field `httpClient'

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -0,0 +1,203 @@
...
+            }
+
+    //  GET /admin/v2/{topic-domain}/{tenant}/{namespace}/{topic}/partitions?checkAllowAutoCreation=true
+    member private this.GetPartitionedTopicMetadataInner (topicName: CompleteTopicName, backoff: Backoff, remainingTimeMs) =
+         backgroundTask {
+            try
+                let randomServiceUri = config.ServiceAddresses[RandomGenerator.Next(0, config.ServiceAddresses.Length)]
+                let topic: string = %topicName
+                let topicRestPath = topic.Replace("persistent://","persistent/").Replace("non-persistent://","non-persistent/")
+                let! response = clientCnx.GetStreamAsync (randomServiceUri.AbsoluteUri + $"admin/v2/%s{%topicRestPath}/partitions" + "?checkAllowAutoCreation=true")
```
Lanayx: I think, `%topicRestPath` can be simplified to `topicRestPath`, here and in other parts of the PR

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -0,0 +1,203 @@
...
+                        MandatoryStop = config.OperationTimeout + config.OperationTimeout
+                        Max = TimeSpan.FromMinutes(1.0)
+                    }
+                let! result = this.GetSchemaInner(topicName, schemaVersion, backoff, int config.OperationTimeout.TotalMilliseconds)
+                return result
+            }
+
+    //  GET /admin/v2/{topic-domain}/{tenant}/{namespace}/{topic}/partitions?checkAllowAutoCreation=true
+    member private this.GetPartitionedTopicMetadataInner (topicName: CompleteTopicName, backoff: Backoff, remainingTimeMs) =
+         backgroundTask {
```
Lanayx: Recursive functions should use async rather than task

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -0,0 +1,203 @@
...
+                      then Uri(brokerResponse.BrokerUrlTls)
+                      else Uri(brokerResponse.BrokerUrl)
+            let resultEndpoint = DnsEndPoint(uri.Host, uri.Port)
+            return { LogicalAddress = LogicalAddress resultEndpoint; PhysicalAddress = PhysicalAddress resultEndpoint }
+        }
+
+    //  GET /admin/v2/namespaces/{tenant}/{namespace}/topics?mode=PERSISTENT
+    //  GET /admin/v2/namespaces/{tenant}/{namespace}/topics?mode=NON_PERSISTENT
+    member private this.GetTopicsUnderNamespaceInner (ns: NamespaceName, backoff: Backoff, remainingTimeMs: int, isPersistent: bool) =
+        backgroundTask {
```
Lanayx: Recursive functions should use async rather than task

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -0,0 +1,203 @@
...
+                Log.Logger.LogWarning(ex, "GetTopicsUnderNamespace failed will retry in {0} ms", delay)
+                do! Async.Sleep delay
+                return! this.GetTopicsUnderNamespaceInner(ns, backoff, remainingTimeMs - delay, isPersistent)
+        }
+
+    //  GET /admin/v2/schemas/{tenant}/{namespace}/{topic}/schema
+    //  GET /admin/v2/schemas/{tenant}/{namespace}/{topic}/schema/{version}
+    member private this.GetSchemaInner(topicName: CompleteTopicName, schemaVersion: SchemaVersion option,
+                              backoff: Backoff, remainingTimeMs: int) =
+         backgroundTask {
```
Lanayx: Recursive functions should use async rather than task

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -0,0 +1,203 @@
...
+                    Name = topicRestPath
+                    Type = schemaResponse.Type
+                    Properties = schemaResponse.Properties
+                    Schema = SchemaData
+                }
+                let topicSchema: TopicSchema = {
+                    SchemaVersion = Some schemaVersion
+                    SchemaInfo = schemaInfo
+                }
+                return Some topicSchema
```
Lanayx, line 199: Binary GetSchema method expects cases when schema can be None, but for Http schema, it's always Some, why is it so?

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -0,0 +1,203 @@
...
+                let delay = Math.Min(backoff.Next(), remainingTimeMs)
+                if delay <= 0 then
+                    raise (TimeoutException "Could not GetSchema within configured timeout.")
+                Log.Logger.LogWarning(ex, "GetSchema failed will retry in {0} ms", delay)
+                do! Async.Sleep delay
+                return! this.GetSchemaInner(topicName, schemaVersion, backoff, remainingTimeMs - delay)
+        }
+
+    //  Convert the key/value schema info data json bytes to key/value schema info data bytes.
+    member private this.ConvertKeyValueDataStringToSchemaInfoSchema(keyValueDataString: string) =
```
Lanayx: This name seems to be over-complicated to me, probably AI-generated. Can you please simplify this and also remove non-applicable comments

tests/IntegrationTests/HttpLookupService.fs
```diff
@@ -0,0 +1,184 @@
...
+
+[<CLIMutable>]
+type SimpleRecord =
+    {
+        Name: string
+        Age: int
+    }
+
+[<CLIMutable>]
+type IncompatibleRecord =
```
Lanayx: Please change this record, since you are not testing incompatibility

Lanayx (changes requested):
Hi, thank you for this giant piece of work. I've added first portion of comments, might add some more a bit later

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: fixed

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: fixed

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: fixed

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: fixed

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: fixed

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: fixed
I copyed this function name from java sdk for consistency, but I also think it is a little bit over-complicated.
https://github.com/apache/pulsar/blob/998bb51f5ea256c59110e5b8bef0f9daef1ac48d/pulsar-client/src/main/java/org/apache/pulsar/client/impl/HttpLookupService.java#L228

tests/IntegrationTests/HttpLookupService.fs
geniusjoe: fixed
I change this name from `IncompatibleRecord`  to `ValueSchema`

src/Pulsar.Client/Common/DTO.fs
geniusjoe: I use `Broker` type as the return type in public interface `ILookupService.GetBroker: CompleteTopicName -> Task<Broker>`, which need to be public accessed.

src/Pulsar.Client/Common/DTO.fs
geniusjoe: I use this type in `HttpLookupService.GetBrokerInner() JsonSerializer.Deserialize<HttpLookupTopicResult>` function. When I set type to internal, I encountered below exception:
```
C:\Users\ninjazhou\AppData\Local\Programs\Rider\plugins\dpa\DotFiles\JetBrains.DPA.Runner.exe --handle=26084 --backend-pid=56200 --etw-collect-flags=67108622 --detach-event-name=dpa.detach.56200.229 --refresh-interval=1 -- C:\Users\ninjazhou\.dotnet\dotnet.exe C:/Users/ninjazhou/Documents/code/githubProjects/pulsar-client-dotnet/tests/IntegrationTests/bin/Debug/net8.0/IntegrationTests.dll --filter-test-list "HttpLookupService"
[11:12:12 INF] EXPECTO? Running tests... <Expecto>
[11:12:13.264 WRN 9] producer(1, concurrent, -1) ConnectionHandler Error reconnecting to persistent://public/default/topic-8683c357746c4b11afe1165cc14dd409 Current state Uninitialized
System.NotSupportedException: Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type 'Pulsar.Client.Common.HttpLookupTopicResult'. Path: $ | LineNumber: 0 | BytePositionInLine: 1.
 ---> System.NotSupportedException: Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type 'Pulsar.Client.Common.HttpLookupTopicResult'.
   --- End of inner exception stack trace ---
   at System.Text.Json.ThrowHelper.ThrowNotSupportedException(ReadStack& state, Utf8JsonReader& reader, NotSupportedException ex)
   at System.Text.Json.Serialization.Converters.ObjectDefaultConverter`1.OnTryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value)
   at System.Text.Json.Serialization.JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
```
It seems that `JsonSerializer.Deserialize` needs a public type to construct its result, should we leave these type to public?

src/Pulsar.Client/Common/DTO.fs
geniusjoe: Similar to the above reason, I use this type in `this.GetSchemaInner() JsonSerializer.Deserialize<HttpSchemaInfo>`, which maybe needs a public type.

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: Yes in httpLookupService this topicSchema it's always Some. 
Class member `GetSchemaInner()` is aimed for realizing `ILookupService GetSchema: CompleteTopicName * ?schema: SchemaVersion -> Task<TopicSchema option>` interface, and I need to add a `Some` wrapper to pass compiler `TopicSchema option` type checking.

Lanayx: Right, I wondered about Pulsar perspective - in theory there should be no logical differences between schemes, if it can be None in "pulsar" scheme, there should be a usecase where it's None in "http" scheme as well.

geniusjoe: @Lanayx Yes you are right. I checked pulsar java sdk `getSchema()` realization, it will return a `HTTP_NOT_FOUND` StatusCode when schema does not exist. And I fixed this issue in the latest commit revision.
https://github.com/apache/pulsar/blob/998bb51f5ea256c59110e5b8bef0f9daef1ac48d/pulsar-client/src/main/java/org/apache/pulsar/client/impl/HttpLookupService.java#L239

src/Pulsar.Client/Common/DTO.fs
Lanayx: Does ILookupService have to be public?

src/Pulsar.Client/Common/DTO.fs
Lanayx: Have you tried annotating the type with `[<CLIMutable>]` attribute?

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -0,0 +1,217 @@
...
+                let brokerResponse = JsonSerializer.Deserialize<PartitionedTopicMetadata>(response, jsonOptions)
+                return brokerResponse
+
+            with Flatten ex ->
+                let nextDelay = Math.Min(backoff.Next(), remainingTimeMs)
+                if nextDelay <= 0 then
+                    reraize ex
+                Log.Logger.LogWarning(ex, "GetPartitionedTopicMetadata failed will retry in {0} ms", nextDelay)
+                do! Async.Sleep nextDelay
+                return! this.GetPartitionedTopicMetadataInner(topicName, backoff, remainingTimeMs - nextDelay)
```
Lanayx, line 106: Here and in other places - do we want to retry any error? Does java client retry every error possible?

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -0,0 +1,217 @@
...
+open Pulsar.Client.Schema
+
+type internal HttpLookupService (config: PulsarClientConfiguration, _connectionPool: ConnectionPool) =
+
+    let httpClient = new HttpClient(new SocketsHttpHandler(
+        PooledConnectionLifetime = TimeSpan.FromSeconds(30),
+        AllowAutoRedirect = true
+    ))
+    let jsonOptions = JsonSerializerOptions(
+        PropertyNameCaseInsensitive = true
```
Lanayx: I'm not sure this is the best solution. If the fields are lowercase, it's better to just use lower-case strategy than case insensitive or just add `[<JsonPropertyName>]` attributes over the properties (like in AuthenticationOauth2.fs)

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -0,0 +1,217 @@
...
+                }
+                let topicSchema: TopicSchema = {
+                    SchemaVersion = Some schemaVersion
+                    SchemaInfo = schemaInfo
+                }
+                return Some topicSchema
+
+            with
+                | :? HttpRequestException as ex ->
+                    if ex.StatusCode = Nullable HttpStatusCode.NotFound then
```
Lanayx: I think this condition can be included into match (using when), so the execution in negative case will fall into the other branch, where no reraize is done

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -0,0 +1,217 @@
...
+open System.Net
+open FSharp.UMX
+open System.Text.Json
+open Microsoft.Extensions.Logging
+open Pulsar.Client.Schema
+
+type internal HttpLookupService (config: PulsarClientConfiguration, _connectionPool: ConnectionPool) =
+
+    let httpClient = new HttpClient(new SocketsHttpHandler(
+        PooledConnectionLifetime = TimeSpan.FromSeconds(30),
```
Lanayx: I suggest we keep it at 2 minutes as suggested in docs
https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#recommended-use

src/Pulsar.Client/Common/DTO.fs
geniusjoe: I think this interface could be set to internal.
I change `ILookupService` from public to internal, and move this interface to `src/Pulsar.Client/Internal/ILookupService.fs`
Currently below types are able to set to internal and I changed them in my latest commit:  
```
TopicSchema
LogicalAddress
Broker
```

src/Pulsar.Client/Common/DTO.fs
geniusjoe: I set `HttpLookupTopicResult` type to `[<CLIMutable>]` like below:
```
[<CLIMutable>]
type internal HttpLookupTopicResult =
    {
        BrokerUrl : string
        BrokerUrlTls: string
        HttpUrl: string
        HttpUrlTls: string
    }        
```
But it seems that client will still throw `System.NotSupportedException`:
```
 C:\Users\ninjazhou\AppData\Local\Programs\Rider\plugins\dpa\DotFiles\JetBrains.DPA.Runner.exe --handle=8760 --backend-pid=14552 --etw-collect-flags=67108622 --detach-event-name=dpa.detach.14552.4 --refresh-interval=1 -- C:\Users\ninjazhou\.dotnet\dotnet.exe C:/Users/ninjazhou/Documents/code/githubProjects/pulsar-client-dotnet/tests/IntegrationTests/bin/Debug/net8.0/IntegrationTests.dll --filter-test-list "HttpLookupService"
[13:14:29 INF] EXPECTO? Running tests... <Expecto>
[13:14:30.209 WRN 16] producer(1, , -1) ConnectionHandler Error reconnecting to persistent://public/default/ae9ad88d67874ec0b7f34c236041b07d-1 Current state Uninitialized
System.NotSupportedException: Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type 'Pulsar.Client.Common.HttpLookupTopicResult'. Path: $ | LineNumber: 0 | BytePositionInLine: 1.
 ---> System.NotSupportedException: Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type 'Pulsar.Client.Common.HttpLookupTopicResult'.
   --- End of inner exception stack trace ---
   at System.Text.Json.ThrowHelper.ThrowNotSupportedException(ReadStack& state, Utf8JsonReader& reader, NotSupportedException ex)
   at System.Text.Json.Serialization.Converters.ObjectDefaultConverter`1.OnTryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value)
   at System.Text.Json.Serialization.JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
```

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: In fact these api return fields as lower camel case format. I set  `[<JsonPropertyName>]` in below types:
```
PartitionedTopicMetadata
HttpLookupTopicResult
HttpSchemaInfo
```

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: Fixed

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: There actually has a bug in my previous code. When an exception is instance of `HttpRequestException` but its status code not equal to `HttpStatusCode.NotFound`, the exception will be throwed to caller chain regardless of backoff policy.
I fix this in my latest commit, please review if my changes are correct.

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: Java client has a retry strategy in `org/apache/pulsar/client/impl/ConnectionHandler.reconnectLater()` to call getBroker() function, which will retry every error.
https://github.com/apache/pulsar/blob/966a396007aba232357170954fb866e50169bdd6/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConnectionHandler.java#L164

I think other lookup functions like `getSchema()`  in java don't have retry strategy, but it is better to use one. When we rolling update broker cluster, or one broker crash for some reason, the lookup service can still pick another broker service address from serviceUrls to reach broker cluster in retry strategy.

src/Pulsar.Client/Common/DTO.fs
Lanayx: Ok, last try :)

Since those Http types are one-off, I guess we could use anonymous records instead, so for example instead of HttpLookupTopicResult it should be possible to use this type for deserialization:
```fsharp
{|  brokerUrl : string; brokerUrlTls: string; httpUrl: string; httpUrlTls: string |}
```

geniusjoe: @Lanayx 
Yes I tested and it can work. It seems that till now Json deserializer cannot parse private type class for reflect performance, but it can parse anonymous record.
I deleted httpLookup related public type classes in DTO.fs and added `PropertyNamingPolicy = JsonNamingPolicy.CamelCase` deserialize option in my latest commit.

https://github.com/fsprojects/FSharp.Json/issues/33

https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/anonymous-records#anonymous-records-are-serializable

src/Pulsar.Client/Common/DTO.fs
geniusjoe: Use anonymous record as an alternative to private type class.

Lanayx (approved):
LGTM

Issue #299, opened 2025-03-06
Support authentication in http lookUpService


geniusjoe:
**Is your feature request related to a problem? Please describe.**
According to pulsar [document](https://pulsar.apache.org/docs/next/security-overview/#how-authentication-works-in-pulsar
), http lookup service needs to pass properties in http header when using authentication :

> When using HTTP/HTTPS protocol for cluster management, each client passes the authentication data based on the HTTP/HTTPS request header, and brokers check the data upon request.

**Describe the solution you'd like**
I think it is not difficult to realize because we have already accomplished AuthenticationToken dataProvider, only we need to do is to set token in http headers:
1. Refer to [Java realization](https://github.com/apache/pulsar/blob/0cc266d54d69205c78dc1bcf03f0b608c20fb62b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/HttpClient.java#L181), we should add function `hasDataForHttp()` and `newRequestHeader()` in `AuthenticationDataProvider` type to generate http headers from token
2. Create a type class called `HttpClient`, this class will initialize authentication related members and expose a [public function](https://github.com/apache/pulsar/blob/0cc266d54d69205c78dc1bcf03f0b608c20fb62b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/HttpClient.java#L151) `Get()` to wrap authentication params
3. Replace `httpClient.GetStreamAsync()` with `httpClient.Get()` in `HttpLookupService` to enable authentication

**Additional context**
Golang realization 
https://github.com/apache/pulsar-client-go/blob/020b55c1dce4475fe8781f4fb6f858de903844b8/pulsaradmin/pkg/admin/auth/token.go#L86

https://github.com/apache/pulsar-client-go/blob/020b55c1dce4475fe8781f4fb6f858de903844b8/pulsar/internal/http_client.go#L78

Lanayx:
Approved, one thing that I'd change is to call it PulsarHttpClient to avoid naming clash with native HttpClient

Lanayx:
Added in 3.8.1

PR #300, opened 2025-03-06
Support http lookup authentication(oauth2 and token)

geniusjoe:
support https://github.com/fsprojects/pulsar-client-dotnet/issues/299

# Changes
- Add function `HasDataForHttp()` and `GetHttpHeaders()` in AuthenticationDataProvider type to generate http headers authentication properties
- Create a type class called PulsarHttpClient, this class will initialize authentication related members and expose a `GetStreamAsync()` function
- Replace `httpClient.GetStreamAsync()` with `pulsarHttpClient.GetStreamAsync()` in HttpLookupService to enable authentication

### DIFF
--- a/src/Pulsar.Client/Internal/HttpLookupService.fs
+++ b/src/Pulsar.Client/Internal/HttpLookupService.fs
@@ -3,7 +3,6 @@
 open System.Collections.Generic
 open System.IO
 open System.Net.Http
-open System.Text.Json.Serialization
 open Pulsar.Client.Api
 open Pulsar.Client.Common
 open System
@@ -15,14 +14,7 @@ open Pulsar.Client.Schema
 
 type internal HttpLookupService (config: PulsarClientConfiguration, _connectionPool: ConnectionPool) =
 
-    let httpClient = new HttpClient(new SocketsHttpHandler(
-        PooledConnectionLifetime = TimeSpan.FromMinutes(2),
-        AllowAutoRedirect = true
-    ))
-    let jsonOptions = JsonSerializerOptions(
-        PropertyNamingPolicy = JsonNamingPolicy.CamelCase
-     )
-    do jsonOptions.Converters.Add(JsonStringEnumConverter())
+    let pulsarHttpClient = PulsarHttpClient(config)
 
     interface ILookupService with
 
@@ -57,8 +49,25 @@ type internal HttpLookupService (config: PulsarClientConfiguration, _connectionP
                 return result
             }
 
+        //  GET /lookup/v2/topic/{topic-domain}/{tenant}/{namespace}/{topic}
         member this.GetBroker(topicName : CompleteTopicName) =
-            this.GetBrokerInner(topicName)
+            backgroundTask {
+                let randomServiceUri = config.ServiceAddresses[RandomGenerator.Next(0, config.ServiceAddresses.Length)]
+                let topic: string = %topicName
+                let topicRestPath = topic.Replace("persistent://","persistent/").Replace("non-persistent://","non-persistent/")
+                let! brokerResponse =
+                    randomServiceUri.AbsoluteUri + $"lookup/v2/topic/%s{topicRestPath}"
+                    |> pulsarHttpClient.Get<{| BrokerUrl : string;
+                                               BrokerUrlTls: string;
+                                               HttpUrl: string;
+                                               HttpUrlTls: string |}>
+                let uri = if config.UseTls then
+                            Uri(brokerResponse.BrokerUrlTls)
+                          else
+                            Uri(brokerResponse.BrokerUrl)
+                let resultEndpoint = DnsEndPoint(uri.Host, uri.Port)
+                return { LogicalAddress = LogicalAddress resultEndpoint; PhysicalAddress = PhysicalAddress resultEndpoint }
+            }
 
         member this.GetTopicsUnderNamespace (ns: NamespaceName, isPersistent: bool) =
             backgroundTask {
@@ -91,11 +100,10 @@ type internal HttpLookupService (config: PulsarClientConfiguration, _connectionP
                 let randomServiceUri = config.ServiceAddresses[RandomGenerator.Next(0, config.ServiceAddresses.Length)]
                 let topic: string = %topicName
                 let topicRestPath = topic.Replace("persistent://","persistent/").Replace("non-persistent://","non-persistent/")
-                let! response =
+                let! brokerResponse =
                     randomServiceUri.AbsoluteUri + $"admin/v2/%s{topicRestPath}/partitions?checkAllowAutoCreation=true"
-                    |> httpClient.GetStreamAsync
+                    |> pulsarHttpClient.Get<{| Partitions: int |}>
                     |> Async.AwaitTask
-                let brokerResponse = JsonSerializer.Deserialize<{| Partitions: int |}>(response, jsonOptions)
                 return { Partitions = brokerResponse.Partitions }
 
             with Flatten ex ->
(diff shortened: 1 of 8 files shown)


src/Pulsar.Client/Internal/PulsarHttpClient.fs
```diff
@@ -0,0 +1,34 @@
...
+    ))
+
+    member internal this.GetStreamAsync (requestUri: string) =
+        backgroundTask {
+            if authenticationDataProvider.HasDataForHttp() then
+                let request = new HttpRequestMessage(HttpMethod.Get, requestUri)
+                authenticationDataProvider.GetHttpHeaders() |> Seq.iter (
+                    fun headerProperty -> request.Headers.Add(headerProperty.Key, headerProperty.Value)
+                )
+                let! response = httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead) |> Async.AwaitTask
```
Lanayx: Async.AwaitTask is not needed here

src/Pulsar.Client/Internal/PulsarHttpClient.fs
```diff
@@ -0,0 +1,34 @@
...
+    let httpClient = new HttpClient(new SocketsHttpHandler(
+        PooledConnectionLifetime = TimeSpan.FromMinutes(2),
+        AllowAutoRedirect = true
+    ))
+
+    member internal this.GetStreamAsync (requestUri: string) =
+        backgroundTask {
+            if authenticationDataProvider.HasDataForHttp() then
+                let request = new HttpRequestMessage(HttpMethod.Get, requestUri)
+                authenticationDataProvider.GetHttpHeaders() |> Seq.iter (
```
Lanayx: Please rewrite this into for loop for readability

src/Pulsar.Client/Internal/PulsarHttpClient.fs
Lanayx: Also, why is it `HttpCompletionOption.ResponseHeadersRead`?

src/Pulsar.Client/Internal/PulsarHttpClient.fs
```diff
@@ -0,0 +1,34 @@
+﻿namespace Pulsar.Client.Internal
+
+open System.Net.Http
+open Pulsar.Client.Api
+open System
+
+//  This class is mainly aimed for http lookup service
```
Lanayx: Can you please add a comment, that in Java it's just HttpClient

src/Pulsar.Client/Internal/PulsarHttpClient.fs
```diff
@@ -0,0 +1,34 @@
...
+type internal PulsarHttpClient (config: PulsarClientConfiguration) =
+
+    let authenticationDataProvider = config.Authentication.GetAuthData()
+
+    let httpClient = new HttpClient(new SocketsHttpHandler(
+        PooledConnectionLifetime = TimeSpan.FromMinutes(2),
+        AllowAutoRedirect = true
+    ))
+
+    member internal this.GetStreamAsync (requestUri: string) =
```
Lanayx: Since type is internal, no need to specify internal on members

src/Pulsar.Client/Internal/PulsarHttpClient.fs
geniusjoe: > Async.AwaitTask is not needed here

Fixed

geniusjoe: > Also, why is it `HttpCompletionOption.ResponseHeadersRead`?

I changed `HttpCompletionOption.ResponseHeadersRead` to `HttpCompletionOption.ResponseContentRead` in my latest commit.

I found that `GetStreamAsync()` cannot pass custom header, so I searched StackOverflow and refer this [answer](https://stackoverflow.com/a/47513859/10064702). 

However I think maybe there is no big difference between `HttpCompletionOption.ResponseContentRead` and `HttpCompletionOption.ResponseHeadersRead` in small size http response, because these two options are only determining when to read the response:

> Indicates if [HttpClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-9.0) operations should be considered completed either as soon as a response is available, or after reading the entire response message including the content.

https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpcompletionoption?view=net-9.0

src/Pulsar.Client/Internal/PulsarHttpClient.fs
geniusjoe: Fixed

src/Pulsar.Client/Internal/PulsarHttpClient.fs
geniusjoe: Fixed

src/Pulsar.Client/Internal/PulsarHttpClient.fs
geniusjoe: Fixed

src/Pulsar.Client/Internal/PulsarHttpClient.fs
Lanayx: `ResponseContentRead` is default, so I think it can just be omitted

Lanayx: Actually, I've just realized, that we don't have to have this function return stream. We can make it generic, just like in [Java](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/HttpClient.java#L151) and leverage [GetFromJsonAsync](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.httpclientjsonextensions.getfromjsonasync?view=net-9.0) built-in method inside

geniusjoe: > Actually, I've just realized, that we don't have to have this function return stream. We can make it generic, just like in [Java](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/HttpClient.java#L151) and leverage [GetFromJsonAsync](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.httpclientjsonextensions.getfromjsonasync?view=net-9.0) built-in method inside

@Lanayx 
Fixed, please review again. And I change `GetStreamAsync()` name to `Get()` to make consistent with Java client.

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -113,14 +104,13 @@ type internal HttpLookupService (config: PulsarClientConfiguration, _connectionP
...
-                    HttpUrl: string
-                    HttpUrlTls: string
-                |}>(response, jsonOptions)
+            let! brokerResponse =
+                randomServiceUri.AbsoluteUri + $"lookup/v2/topic/%s{topicRestPath}"
+                |> pulsarHttpClient.Get<{| BrokerUrl : string;
+                                           BrokerUrlTls: string;
+                                           HttpUrl: string;
+                                           HttpUrlTls: string |}>
+                |> Async.AwaitTask
```
Lanayx: Since this method is not recursive/retriable, we don't need async here, Async.AwaitTask as well, and `GetBrokerInner` is not needed, this code can be moved to just `GetBroker`

src/Pulsar.Client/Internal/HttpLookupService.fs
```diff
@@ -140,11 +130,10 @@ type internal HttpLookupService (config: PulsarClientConfiguration, _connectionP
                     match isPersistent with
                     | true -> "PERSISTENT"
                     | false -> "NON_PERSISTENT"
-                let! response =
+                let! brokerResponse =
                     randomServiceUri.AbsoluteUri + $"admin/v2/namespaces/%s{ns.ToString()}/topics?mode=%s{mode}"
-                    |> httpClient.GetStreamAsync
+                    |> pulsarHttpClient.Get<string seq>
```
Lanayx: Can you please change seq to array (in the interface as well)

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: I removed async and Async.AwaitTask in `GetBrokerInner()`, and migrate this code to `GetBroker()` with `background{}` wrapper.

src/Pulsar.Client/Internal/HttpLookupService.fs
geniusjoe: I change `HttpLookupService.fs` `ILookupService` and `BinaryLookupService` `GetTopicsUnderNamespace()` return type. I think this function only used in `PulsarClient.fs.GetTopicsByPattern()` so that we don't need to change caller strategy.

Lanayx (approved):

PR #302, opened 2025-03-25
Split client TLS transport encryption from authentication

RobertIndie:
## Motivation

Bring PIP-158 https://github.com/apache/pulsar/issues/15289 to Pulsar.Client

## Modification

- Introduce a new client configuration: `TlsCertificate`. This combines `tlsKeyFilePath` and `tlsCertificateFilePath` from the Java client. Users need to provide both the certificate and private key in this configuration (the PFX file) for mTLS encryption. If the private key is missing, an exception will be thrown: "TlsCertificate doesn't contain a private key."
  
  - If `useTls` is enabled and TLS authentication data is absent, the client will use `TlsCertificate` as the client certificate for mTLS. This allows users to separate mTLS encryption configuration from authentication configuration, enabling the use of mTLS for encryption and other authentication method like JWT simultaneously.

## Usage

```csharp
var ca = new X509Certificate2(@"/path/to/ca.cert.crt");
var clientCert = new X509Certificate2(@"/path/to/client.pfx");
var client = await new PulsarClientBuilder()
    .ServiceUrl(serviceUrl)
    .TlsTrustCertificate(ca)
    .TlsCertificate(clientCert)
    .Authentication(AuthenticationFactory.Token("token_xxx"))
    .BuildAsync();
```

## Release

We should also cherry-pick this feature to branch 2.x

### DIFF
--- a/src/Pulsar.Client/Internal/ConnectionPool.fs
+++ b/src/Pulsar.Client/Internal/ConnectionPool.fs
@@ -143,7 +143,18 @@ type internal ConnectionPool (config: PulsarClientConfiguration) =
                         if config.UseTls then
                             Log.Logger.LogDebug("Configuring ssl for {0}", physicalAddress)
                             let sslStream = new SslStream(new NetworkStream(socket), false, RemoteCertificateValidationCallback(remoteCertificateValidationCallback))
-                            let clientCertificates = config.Authentication.GetAuthData(physicalAddress.Host).GetTlsCertificates()
+                            let authData = config.Authentication.GetAuthData(physicalAddress.Host)
+                            let clientCertificates =
+                                if authData.HasDataForTls() then
+                                    config.Authentication.GetAuthData(physicalAddress.Host).GetTlsCertificates()
+                                else
+                                    let clientCert = config.TlsCertificate
+                                    if clientCert = null then
+                                        X509Certificate2Collection([||])
+                                    elif not clientCert.HasPrivateKey then
+                                        failwith "TlsCertificate doesn't contain a private key"
+                                    else
+                                        X509Certificate2Collection([| clientCert |])
                             do! sslStream.AuthenticateAsClientAsync(physicalAddress.Host, clientCertificates, config.TlsProtocols, false)
 
                             let pipeConnection = StreamConnection.GetDuplex(sslStream, pipeOptions)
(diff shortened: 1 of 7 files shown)


tests/IntegrationTests/Tls.fs
```diff
@@ -11,44 +11,47 @@ open Pulsar.Client.Common
 open Pulsar.Client.IntegrationTests.Common
 
 
+let tlsTransport client testName =
+    testTask testName {
```
Lanayx, line 16: Can you please add debug logs at the beginning and the end of the test (similar to other integration tests)

tests/IntegrationTests/Tls.fs
```diff
@@ -11,44 +11,47 @@ open Pulsar.Client.Common
 open Pulsar.Client.IntegrationTests.Common
 
 
+let tlsTransport client testName =
+    testTask testName {
+        let client = getSslAdminClient()
```
Lanayx: I think this line should be removed

src/Pulsar.Client/Internal/ConnectionPool.fs
```diff
@@ -143,7 +143,18 @@ type internal ConnectionPool (config: PulsarClientConfiguration) =
...
                             let sslStream = new SslStream(new NetworkStream(socket), false, RemoteCertificateValidationCallback(remoteCertificateValidationCallback))
-                            let clientCertificates = config.Authentication.GetAuthData(physicalAddress.Host).GetTlsCertificates()
+                            let authData = config.Authentication.GetAuthData(physicalAddress.Host)
+                            let clientCertificates =
+                                if authData.HasDataForTls() then
+                                    config.Authentication.GetAuthData(physicalAddress.Host).GetTlsCertificates()
+                                else
+                                    let clientCert = config.TlsCertificate
+                                    if clientCert = null then
+                                        X509Certificate2Collection([||])
```
Lanayx: Round braces can be substituted by space

tests/IntegrationTests/Common.fs
```diff
@@ -29,6 +29,7 @@ let pulsarSslAddress = "pulsar+ssl://127.0.0.1:6651"
 // generate pfx file from pem, leave the password blank
 // openssl pkcs12 -in admin.cert.pem -inkey admin.key-pk8.pem -export -out admin.pfx
 let ca = new Security.Cryptography.X509Certificates.X509Certificate2(@"../ssl/ca.cert.pem")
+let clientCert = new Security.Cryptography.X509Certificates.X509Certificate2(@"../ssl/admin.pfx")
```
Lanayx, line 34: Please add comment why `admin.pfx` can be used here as client certificate

Lanayx (changes requested):

Lanayx (approved):
Thank you merging it in

PR #307, opened 2025-04-08
Use Aegis256 as the symmetric algorithm

RobertIndie:


RobertIndie:
Sorry. I pushed this PR by mistake.

PR #309, opened 2025-04-11
Make consumer DisposeAsync execute in the backgroundTask

RobertIndie:
## Motivation

The `producer.DisposeAsync` runs in the background, while the consumer does not. This can cause inconsistent behavior between closing the producer and the consumer. For example, if a user wants to wait for the result of `consumer.DisposeAsync` using `GetAwaiter`, it may lead to a deadlock if `ConfigureAwait(false)` is not set. It is better to run this dispose in the background to avoid deadlock issues and to ensure consistent behavior with the producer.


Lanayx (approved):
Thank you for noticing this

Lanayx:
@RobertIndie it seems that apachepulsar/pulsar-test-latest-version was deleted recently from docker, so integration tests can't be run, should we use zhaijia/pulsar-test-latest-version? Or maybe ghcr.io/nodece/pulsar-test-latest-version ?

PR #310, opened 2025-04-15
[2.x] Make consumer DisposeAsync execute in the backgroundTask (#309)

RobertIndie:
Cherry-pick: https://github.com/fsprojects/pulsar-client-dotnet/pull/309

## Motivation

The `producer.DisposeAsync` runs in the background, while the consumer does not. This can cause inconsistent behavior between closing the producer and the consumer. For example, if a user wants to wait for the result of `consumer.DisposeAsync` using `GetAwaiter`, it may lead to a deadlock if `ConfigureAwait(false)` is not set. It is better to run this dispose in the background to avoid deadlock issues and to ensure consistent behavior with the producer.

---------

Co-authored-by: Vladimir Shchur <odindafna2006@rambler.ru>
(cherry picked from commit 096bc2356b3e12ff9e77e1451abb6e977e9bc047)

### DIFF
--- a/.github/workflows/dotnetcore-ubuntu.yml
+++ b/.github/workflows/dotnetcore-ubuntu.yml
@@ -5,7 +5,7 @@ on: [pull_request, push]
 jobs:
   build:
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


RobertIndie:
Could you help take a look at the CI error? Seems not related to this PR.

Lanayx:
@RobertIndie Ubuntu image needs to be updated

RobertIndie:
> @RobertIndie Ubuntu image needs to be updated

Thanks. The CI has passed

PR #314, opened 2025-05-18
allow base64 encoded OAuth credentials

DanBiscotti:
Hi, as mentioned in #263 this is a change to allow the same feature on the .NET standard version of the client
Cheers, Dan


Lanayx:
Thank you, published 2.19.0

PR #316, opened 2025-05-21
Retain ordering key on ReconsumeLater

adrianotr:


src/Pulsar.Client/Internal/DeadLetters.fs
```diff
@@ -89,7 +89,7 @@ type internal DeadLetterProcessor<'T>
                     let! rlProducer = rlProducer.Value
                     let key = getOptionalKey message
                     let eventTime = message.EventTime |> Option.ofNullable
-                    let msg = MessageBuilder(message.GetValue(), message.Data, key, propertiesMap, deliverAt, ?eventTime = eventTime)
+                    let msg = MessageBuilder(message.GetValue(), message.Data, key, propertiesMap, deliverAt, orderingKey = message.OrderingKey, ?eventTime = eventTime)
```
Lanayx: This will allow for `Some null`. I think `Option.ofObj` should do the trick (similar to `eventTime`)

tests/IntegrationTests/DeadLetters.fs
```diff
@@ -448,6 +448,44 @@ let tests =
...
+                    .SubscribeAsync()
+
+            let! _ = producer.NewMessage([| 0uy; 1uy; 0uy |], orderingKey = [| 1uy; 0uy; 1uy |]) |> producer.SendAsync
+            let! (msg1 : Message<byte[]>) = consumer.ReceiveAsync()
+            do! consumer.ReconsumeLaterAsync(msg1, %(DateTime.UtcNow.AddSeconds(1.0) |> convertToMsTimestamp))
+            let! (msg2 : Message<byte[]>) = consumer.ReceiveAsync()
+
+            Expect.isNotNull "" msg1.OrderingKey
+            Expect.isNotNull "" msg2.OrderingKey
+            Expect.sequenceEqual "" msg1.OrderingKey msg2.OrderingKey
```
Lanayx: I think that not only keys should be the same between messages, but they should be equal the ordering key sent by producer

Lanayx (changes requested):

src/Pulsar.Client/Internal/DeadLetters.fs
adrianotr: Sorry, still learning F#. I think it's correct now.

Lanayx (approved):

Lanayx:
Thank you, published 3.9.3

PR #319, opened 2025-06-19
Fix pattern consumer fails when a subscribed topic is deleted

RobertIndie:
### Motivation

When a topic that a pattern consumer is subscribed to is deleted, the consumer's sub-consumer tries to reconnect. However, it encounters a "topic not found" error. Currently, this error causes all pending waiters to fail, leading to an `AlreadyClosedException` when `consumer.ReceiveAsync()` is called. This issue can disrupt other consumers.

```
Pulsar.Client.Api.AlreadyClosedException: Consumer is already closed
   at <StartupCode$Pulsar-Client>.$ConsumerImpl.ReceiveWrappedAsync@1442.MoveNext() in xxx/src/Pulsar.Client/Internal/ConsumerImpl.fs:line 1444
   at Pulsar.Client.IntegrationTests.Basic.tests@434-831.Invoke(ResumableStateMachine`1& sm)
   at Pulsar.Client.IntegrationTests.Basic.tests@434-830.Invoke(ResumableStateMachine`1& sm)

```

To fix this, if a sub-consumer encounters a "topic not found" error during reconnection, we should simply remove that consumer from the pattern consumer instead of throwing an exception.

### Modification

- This change prevents pending waiters from failing when a consumer encounters a "topic not found" error during reconnection.

### DIFF
--- a/tests/IntegrationTests/Basic.fs
+++ b/tests/IntegrationTests/Basic.fs
@@ -4,12 +4,14 @@ open System
 open System.Text.Json
 open System.Threading
 open System.Diagnostics
+open System.Net.Http
 
 open Expecto
 open Expecto.Flip
 
 open System.Text
 open System.Threading.Tasks
+open Microsoft.Extensions.Logging
 open Pulsar.Client.Api
 open Pulsar.Client.Common
 open Serilog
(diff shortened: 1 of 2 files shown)


tests/IntegrationTests/Basic.fs
```diff
@@ -373,6 +375,70 @@ let tests =
...
+                
+            let! (producer : IProducer<byte[]>) =
+                client.NewProducer()
+                    .Topic(topicName + "-2")
+                    .CreateAsync()
+                    
+            do! producer.SendAsync([| 1uy |])
+            
+            do! task
+            let (msg1 : Message<byte[]>) = task.Result
```
Lanayx: It should be possible to simplify this with
```fsharp
let! (msg1 : Message<byte[]>) = task
```

tests/IntegrationTests/Basic.fs
```diff
@@ -373,6 +375,70 @@ let tests =
...
+            // Check that the task doesn't fail immediately
+            Expect.isFalse "" task.IsFaulted
+            Expect.isFalse "" task.IsCanceled
+            
+            // Delete topic using HTTP request
+            let deleteUrl = $"{pulsarHttpAddress}/admin/v2/persistent/{topicName}-1?force=true"
+            let! (response: HttpResponseMessage) = commonHttpClient.DeleteAsync(deleteUrl)
+            response.EnsureSuccessStatusCode() |> ignore
+            
+            Thread.Sleep 1000 // This make sure that the topic won't be recreated
```
Lanayx: It's better to use
```fsharp
do! Task.Delay(1000)
```

tests/IntegrationTests/Basic.fs
```diff
@@ -373,6 +375,70 @@ let tests =
             Expect.isTrue "" isReplicated
             Log.Debug("Finished 'Create the replicated subscription should be successful'")
         }
+        
+        testTask "Delete topic subscribed by the pattern consumer should not throw error or recreate topic" {
+            let topicName = "public/default/topic-" + Guid.NewGuid().ToString("N")
```
Lanayx, line 381: Can you also please add `Log.Debug` at the beginning and at the end of test (similar to other tests)? This helps debugging hanging tests

RobertIndie: Yes. I forgot it. Thanks!

Lanayx: At the end too, please

tests/IntegrationTests/Basic.fs
```diff
@@ -373,6 +375,71 @@ let tests =
...
+            Expect.isFalse "" task.IsCompleted
+                
+            let! (producer : IProducer<byte[]>) =
+                client.NewProducer()
+                    .Topic(topicName + "-2")
+                    .CreateAsync()
+                    
+            do! producer.SendAsync([| 1uy |])
+            
+            do! task
```
Lanayx: This line is no longer needed

Lanayx (approved):

PR #326, opened 2025-09-12
OTEL activity key validation and decrease of log level

glebvsmir:
Add validation for activity key to avoid dictionary key not found exception if the OTEL activity or propagators are misconfigured.
Decrease the logging level for consumer and producer OTEL interceptors to avoid verbose warning if OTEL is NOT configured OR misconfigured.


glebvsmir:
Wrong base branch; will reapply commits on top of 2.x

PR #327, opened 2025-09-12
Otel validate key and decrease log level

glebvsmir:


src/Pulsar.Client.Otel/OTelProducerInterceptor.fs
```diff
@@ -42,6 +42,9 @@ type OTelProducerInterceptor<'T>(sourceName: string, log: ILogger) =
 
             if isNull activity then
                 message  //If there are no listeners interested in this activity, the activity above will be null.
+            elif not (mutableDict.ContainsKey activityKey) then
+                log.LogDebug("activityKey {0} is missing in dictionary. Check if OTEL propagators are configured correctly.", activityKey)
```
Lanayx: It looks like this branch will always be chosen, since mutableDict will never contain the key at this point. Also log level should be warning here, since we know that listeners are configured at this point. In other places of this PR reducing to debug makes sense, but not here.

Lanayx (changes requested):
Changes requested

src/Pulsar.Client.Otel/OTelProducerInterceptor.fs
glebvsmir: Agree with comments. Corresponding code changes made and logic moved further down the "activity" branch.

Lanayx (approved):

Lanayx:
I've published 0.1.3, but then noticed a bug in the new implementation, so I had to unlist it. Please create another PR to fix the bug and also confirm it's verified locally before I merged it in

PR #328, opened 2025-09-16
Otel validate key and decrease log level

glebvsmir:
Bug fix for unlisted v0.1.3


glebvsmir:
Stale branch; see PR329 instead

PR #330, opened 2025-09-17
Fix single topic consumer returning message ID with empty topic name

RobertIndie:
## Motivation

Currently, the single topic consumer does not set the topic name for the returned message ID.  
This PR fixes the issue by ensuring that the topic name is correctly set.


Lanayx (approved):
It was actually by design, for some reason I though that TopicName should be empty for non-partitioned topic, but I no longer can recall why. As of today I don't see any objections.

PR #331, opened 2025-09-19
Add producer name attribute to the Message

RobertIndie:
## Motivation

According to the doc: https://pulsar.apache.org/docs/3.0.x/concepts-messaging/#messages, we should also include the `producerName` to the message received by the consumer.


Lanayx:
Build check doesn't pass

Lanayx (approved):

PR #333, opened 2025-10-29
Fix wrong results of hasMessageAvailable after seeking by timestamp

RobertIndie:
## Motivation

Bring this fix https://github.com/apache/pulsar/pull/22363 to the pulsar-client-dotnet

The reader's HasMessageAvailableAsync may return true after a timestamp-based seek, even if there are no available messages in the topic.

### Modifications

Add a boolean flag `HasSoughtByTimestamp` to represent whether the last seek call accepts a timestamp. If it's true, don't take `startMessageId` into comparison with `lastMessageIdInBroker`, just compare the mark-delete position and last position in the GetLastMessageId response.

### DIFF
--- a/src/Pulsar.Client/Internal/ConsumerImpl.fs
+++ b/src/Pulsar.Client/Internal/ConsumerImpl.fs
@@ -97,6 +97,7 @@ type internal ConsumerImpl<'T> (consumerConfig: ConsumerConfiguration<'T>, clien
     let mutable lastMessageIdInBroker = MessageId.Earliest
     let mutable lastDequeuedMessageId = MessageId.Earliest
     let mutable duringSeek = None
+    let mutable hasSoughtByTimestamp = false
     let initialStartMessageId = startMessageId
     let mutable incomingMessagesSize = 0L
     let deadLettersProcessor = consumerConfig.DeadLetterProcessor topicName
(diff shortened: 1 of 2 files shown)


src/Pulsar.Client/Internal/ConsumerImpl.fs
```diff
@@ -1431,6 +1436,10 @@ type internal ConsumerImpl<'T> (consumerConfig: ConsumerConfiguration<'T>, clien
         with get() = Volatile.Read(&lastMessageIdInBroker)
         and private set value = Volatile.Write(&lastMessageIdInBroker, value)
 
+    member this.HasSoughtByTimestamp
+        with get() = Volatile.Read(&hasSoughtByTimestamp)
```
Lanayx: The HasSoughtByTimestamp property is unused, so not needed

tests/IntegrationTests/Reader.fs
```diff
@@ -272,6 +272,85 @@ let tests =
...
+                        Expect.isTrue "should have message" hasMessage
+                    else
+                        let! hasMessage = reader.HasMessageAvailableAsync()
+                        Expect.isFalse "should not have message" hasMessage
+                
+                let futureTimestamp = %(DateTimeOffset.UtcNow.AddMinutes(1.0).ToUnixTimeMilliseconds())
+                
+                // The Seek operation does not implement a backoff mechanism. It will fail if the connection is not
+                // ready, so wait for a short period until the connection becomes available.
+                do! Async.Sleep(1000)
```
Lanayx: Please use Task.Delay here and below

Lanayx (approved):

Lanayx:
Thank you, released 3.12.1

PR #335, opened 2025-11-08
Integration Tests for BatchReceiveAsync

Ttpetrov:
Environment used .Net 8.0 C#/Pulsar Server4.0.0

Summary: Added two new integration tests for batch receive functionality to verify correct behavior with message backlogs.
Tests Added:
1.	"Batch receive respects max size with large backlog" ✅
•	Verifies that BatchReceiveAsync() respects the configured maxBatchSize (8 messages) when consuming from a large pre-existing backlog (2000 messages)
•	Produces all messages BEFORE consumer subscription to ensure backlog exists
•	Validates that no batch exceeds the configured maximum size
2.	"Batch receive respects max size with large backlog and concurrent production" ❌ (Expected to Fail)
•	This test is currently failing and is EXPECTED to fail
•	Demonstrates a potential bug/limitation where batch size limits are violated when:
•	A large backlog exists (2000 messages)
•	Additional messages are continuously produced during consumption (1000 more messages)
•	Consumer has simulated processing delays (500ms per batch)
•	The failure reveals that concurrent production + backlog may cause BatchReceiveAsync() to return batches larger than the configured maxBatchSize
Purpose: The second test documents the reproduction case for investigation and future fix. It should remain as a failing test until the underlying issue is resolved.

### DIFF
--- a/tests/IntegrationTests/Batching.fs
+++ b/tests/IntegrationTests/Batching.fs
@@ -368,7 +368,7 @@ let tests =
             let backlogMessagesCount = 2000
             let maxBatchSize = 8
             let batchTimeout = TimeSpan.FromSeconds(5.0)
-            let topicName = "public/ods_test/test-topic"
+            let topicName = "public/default/topic-" + Guid.NewGuid().ToString("N")
 
             let! producer =
                 client.NewProducer()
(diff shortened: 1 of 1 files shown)


Lanayx:
Thank you for an example of the failing test! I was able to locate the problem and cover it with much simpler unit tests [here ](https://github.com/fsprojects/pulsar-client-dotnet/commit/5b562ca68d2a092364e7d9d3390c4a75a02edc9b)

Issue #340, opened 2026-01-19
Unexpected reading behaviour after seeking by time and using `HasMessageAvailableAsync`


calindima:
Pulsar client version: 3.13.1
Pulsar cluster version: 4.0.7

There's some unexpected behaviour when starting a new Reader client, seeking by timestamp and then trying to check if there are any messages available to read.

Calling `reader.HasMessageAvailableAsync()` before the call to `reader.ReadNextAsync()` when previously seeking by a certain timestamp seems to result in the cursor moving at the end of the topic and the reader client hanging until there's new messages being produced.

The following results in the reader client receiving only messages produced after the call to `ReadNextAsync` instead of everything after the `timestamp`:
```c#
        await reader.SeekAsync(timestamp);

         while (await reader.HasMessageAvailableAsync())
        {
            var msg = await reader.ReadNextAsync(CancellationToken.None);

            logger.Log(LogLevel.Information,
                $"Received message. PublishTime: {DateTimeOffset.FromUnixTimeMilliseconds(msg.PublishTime)} Payload: {Encoding.UTF8.GetString(msg.Data)}");
        }
```

The following results in the reader client behaving as expected:
```c#
        await reader.SeekAsync(timestamp);

        do
        {
            var msg = await reader.ReadNextAsync();

            logger.Log(LogLevel.Information,
                $"Received message. PublishTime: {DateTimeOffset.FromUnixTimeMilliseconds(msg.PublishTime)} Payload: {Encoding.UTF8.GetString(msg.Data)}");
        } while (await reader.HasMessageAvailableAsync());
```

Lanayx:
@calindima It looks like you are hitting [the expected case](https://github.com/fsprojects/pulsar-client-dotnet/blob/develop/src/Pulsar.Client/Internal/ConsumerImpl.fs#L1138-L1142), @RobertIndie do you happen to know is it by design or not?

RobertIndie:
> [@calindima](https://github.com/calindima) It looks like you are hitting [the expected case](https://github.com/fsprojects/pulsar-client-dotnet/blob/develop/src/Pulsar.Client/Internal/ConsumerImpl.fs#L1138-L1142), [@RobertIndie](https://github.com/RobertIndie) do you happen to know is it by design or not?

This is an unexpected issue. I pushed a PR to fix it: https://github.com/fsprojects/pulsar-client-dotnet/pull/344

Lanayx:
Should be fixed in 3.13.3

PR #342, opened 2026-01-28
[2.x] Fix wrong behavior of hasMessageAvailable after seeking by timestamp

RobertIndie:
## Motivation

Cherry-pick https://github.com/fsprojects/pulsar-client-dotnet/pull/333 to 2.x branch.

The reader's HasMessageAvailableAsync may return true after a timestamp-based seek, even if there are no available messages in the topic.

### Modifications

Add a boolean flag `HasSoughtByTimestamp` to represent whether the last seek call accepts a timestamp. If it's true, don't take `startMessageId` into comparison with `lastMessageIdInBroker`, just compare the mark-delete position and last position in the GetLastMessageId response.

### DIFF
--- a/src/Pulsar.Client/Internal/ConsumerImpl.fs
+++ b/src/Pulsar.Client/Internal/ConsumerImpl.fs
@@ -96,6 +96,7 @@ type internal ConsumerImpl<'T> (consumerConfig: ConsumerConfiguration<'T>, clien
     let mutable lastMessageIdInBroker = MessageId.Earliest
     let mutable lastDequeuedMessageId = MessageId.Earliest
     let mutable duringSeek = None
+    let mutable hasSoughtByTimestamp = false
     let initialStartMessageId = startMessageId
     let mutable incomingMessagesSize = 0L
     let receiverQueueRefillThreshold = consumerConfig.ReceiverQueueSize / 2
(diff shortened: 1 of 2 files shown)


RobertIndie:
The CI is failed related to this issue: https://github.com/fsprojects/pulsar-client-dotnet/issues/343. Could you help try rerun it? @Lanayx

Lanayx:
@RobertIndie Before we do the backport can you please comment on the two issues
1) This ticket https://github.com/fsprojects/pulsar-client-dotnet/issues/340
2) I've recently realized that `hasSoughtByTimestamp` never returns back to false, once set to `true` it stays there forever. Shouldn't it be reset after subscription is reestablished?

Lanayx:
> The CI is failed related to this issue: #343. Could you help try rerun it? @Lanayx

Done

RobertIndie:
> @RobertIndie Before we do the backport can you please comment on the two issues
> 
> 1. This ticket [Unexpected reading behaviour after seeking by time and using `HasMessageAvailableAsync` #340](https://github.com/fsprojects/pulsar-client-dotnet/issues/340)
> 2. I've recently realized that `hasSoughtByTimestamp` never returns back to false, once set to `true` it stays there forever. Shouldn't it be reset after subscription is reestablished?

Thanks! I pushed a PR to address both these issues: https://github.com/fsprojects/pulsar-client-dotnet/pull/344.

Lanayx:
Thanks for addressing that PR. Do your want to includes those commits in this PR?

As I side note - it looks like we don't need to support 2.x any longer, based on nuget numbers nobody uses latest 2.x packages.
image

RobertIndie:
> Thanks for addressing that PR. Do your want to includes those commits in this PR?

I include that commit to this PR as well now.

> As I side note - it looks like we don't need to support 2.x any longer, based on nuget numbers nobody uses latest 2.x packages.

Some of our users are still relying on version 2.x.

Lanayx:
> Some of our users are still relying on version 2.x.

Right, but nobody uses versions above 2.16.0

PR #359, opened 2026-06-11
Fix batch acknowledgment for deserialized message IDs

MatheusReichert:
See also https://github.com/apache/pulsar/issues/19030 and https://github.com/apache/pulsar/pull/19031

## Motivation

Currently, when a message ID of a batched message is serialized (e.g., stored in a database or outbox envelope) and later deserialized to be acknowledged manually via `AcknowledgeAsync` (or `NegativeAcknowledge`), the acknowledgment fails to propagate to the broker.

This happens because the deserialized `MessageId` gets its own newly instantiated, isolated instance of `BatchMessageAcker`. Since the individual messages in the batch no longer share the same acker instance, calling `AckIndividual` on each isolated acker never drives the outstanding count to `0` for any of the instances. Thus, the consumer never sends the final batch acknowledgment command to the broker, causing all messages in the batch to be redelivered once the acknowledgment timeout expires.

This PR fixes the issue by introducing a `batchAckers` map inside `ConsumerImpl` to track and share `BatchMessageAcker` instances across deserialized message IDs associated with the same batch entry `(LedgerId, EntryId)`.

## Modifications
- Added the `batchAckers` dictionary in `ConsumerImpl.fs` to track active `BatchMessageAcker`s.
- Stored the batch acker in the `batchAckers` map during batch parsing in `ReceiveIndividualMessagesFromBatch`.
- Updated `trySendAcknowledge` and `doTransactionAcknowledgeForResponse` to resolve the shared `BatchMessageAcker` from the map.
- Added clearing of the `batchAckers` dictionary in `closeConsumerTasks` to release resources and prevent memory leaks.
- Added a unit test in `MessageTests.fs` to verify batch message ID acker isolation behavior during serialization and deserialization.

### DIFF
--- a/src/Pulsar.Client/Internal/ConsumerImpl.fs
+++ b/src/Pulsar.Client/Internal/ConsumerImpl.fs
@@ -1,4 +1,4 @@
-﻿namespace Pulsar.Client.Api
+namespace Pulsar.Client.Api
 
 open System.Collections
 
@@ -260,6 +260,8 @@ type internal ConsumerImpl<'T> (consumerConfig: ConsumerConfiguration<'T>, clien
 
     let negativeAcksTracker = NegativeAcksTracker(prefix, consumerConfig.NegativeAckRedeliveryDelay, negativeAcksRedeliver)
 
+    let batchAckers = Dictionary<struct(LedgerId*EntryId), BatchMessageAcker>()
+
     let getConnectionState() = connectionHandler.ConnectionState
     let sendAckPayload (cnx: ClientCnx) payload = cnx.Send payload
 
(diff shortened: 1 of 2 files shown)


Lanayx:
Hi, thanks for raising the issue, however given the referenced Java issue and unmerged PR, I'm not sure that this should be fixed at all, it rather looks like unsupported case by pulsar.
As far as I know, the main use case for saving MessageId before consumer restart to start from it later, so the byte structure was designed particuarly for that. It doesn't seem like there is any other pulsar client that went above that.

Lanayx:
@RobertIndie Do you have any insights on this issue?

PR #367, opened 2026-09-09
Fix producer sequence ordering for delayed messages and key-based batches

RobertIndie:
## Motivation

With broker deduplication enabled, two producer paths can submit sequence IDs out of order and cause `PersistenceError: Cannot determine whether the message is a duplicate at this time`, connection recovery, and successful send tasks for payloads that were skipped as duplicates.

First, an ordinary message can remain buffered while a later message with `deliverAt` is sent immediately. The broker sees the larger sequence ID first. Second, KeyBased batching can turn A(0), B(1), A(2), B(3) into overlapping send ranges A:[0,2] and B:[1,3]. Accepting A advances producer-wide deduplication progress past B's starting ID.

## Modification

- Submit the accumulated batch before handling a message that bypasses batching. This preserves submission order without waiting for the batch ACK.
- Follow the Java client's `BatchMessageKeyBasedContainer.createOpSendMsgs()` approach: use each key batch's highest sequence ID as its send ID, then sort key batches by that ID before serialization. For A(0), B(1), B(2), A(3), send B with ID 2 before A with ID 3. 
Java related code:
https://github.com/apache/pulsar/blob/9dced41ea515f7534c1861cdd85a5e21b3d790c8/pulsar-client/src/main/java/org/apache/pulsar/client/impl/BatchMessageKeyBasedContainer.java#L99-L108
- Name the internal batch wrapper field `SequenceId` and use it consistently in outer metadata, SEND commands, and pending ACK tracking. KeyBased SEND sequence and highest-sequence values are equal; individual messages retain their original sequence IDs, keys, payloads, and batch indexes. Default batch numbering retains its existing semantics.


Lanayx:
@copilot resolve the merge conflicts in this pull request

PR #368, opened 2026-09-16
Fix BatchReceiveAsync never delivering a message larger than MaxNumBytes

SlavchoGeorgiev:
## Problem

`Messages.CanAdd` (`src/Pulsar.Client/Common/Message.fs`) rejects a message whose payload alone exceeds `BatchReceivePolicy.MaxNumBytes`, even when the batch is still **empty**.

Because `hasEnoughMessagesForBatchReceive` (`Common/ConsumerBase.fs`) is true as soon as `incomingMessagesSize >= MaxNumBytes`, every `BatchReceiveAsync` call then replies **synchronously with an empty batch**, forever:

1. the oversize message stays at the head of `incomingMessages`,
2. it is never delivered to the application and never acknowledged, so it blocks the subscription (and is redelivered to the next consumer on failover),
3. an application looping on `BatchReceiveAsync` spins at full CPU with a GC storm, with nothing logged.

The Java client explicitly allows this case in `MessagesImpl.canAdd`:

```java
if (currentNumberOfMessages == 0) {
    // It's ok to add at least one message into a batch.
    return true;
}
```

We hit this in production with a `BatchReceivePolicy(500, 10 MiB, 1 ms)` consumer and a ~10.5 MB (uncompressed) message: the pod that received it stopped consuming and burned ~1.4 cores at 45–95 Gen0 GCs/s until it was killed, after which the message was redelivered to the next pod.

## Fix

`CanAdd` now always accepts the first message of a batch, then applies the count and size limits as before. Both `ConsumerImpl.replyWithBatch` and `MultiTopicsConsumerImpl.replyWithBatch` go through `CanAdd`, so partitioned and multi-topic consumers are covered too.

## Test

New integration test in `tests/IntegrationTests/Batching.fs`: **"Batch receive delivers a single message larger than MaxNumBytes"** — a 1 MB message against a 512 KB `MaxNumBytes`.

- Before the fix: errors after 10 s of continuous empty batches (`TaskCanceledException` on the cancellation token).
- After the fix: passes, and the existing "Batch receive works with regular consumer" still passes.
- Unit tests: 201 passed.

Run against the standalone compose broker with:

```
dotnet run --project tests/IntegrationTests -c Release -p:DefineConstants=NOTLS -- --filter-test-case "Batch receive"
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Lanayx (approved):

Lanayx:
Thank you for the PR and explanation!

PR #369, opened 2026-09-17
Close the consumer at the broker when its mailbox fails

SlavchoGeorgiev:
## What

When an unhandled exception hits the consumer's mailbox, the client marks the consumer `Failed` and stops it locally, but never sends `CloseConsumer` to the broker.

## Why it matters

The broker still thinks the dead consumer is attached:
- On an `Exclusive` subscription, no other consumer can take its place (`ConsumerBusyException`) until the process exits.
- On a `Failover` subscription, the broker keeps the dead consumer as the active one, so the subscription stalls until the pod is killed.

## Fix

On a mailbox fault, if the connection is still `Ready`, remove the consumer from the connection's registry and send `CloseConsumer` before transitioning to `Failed`, so the broker frees the subscription right away.

## Testing

Added an integration test in `Basic.fs` that force-crashes a consumer's mailbox and asserts a second consumer can immediately subscribe Exclusively. It fails on `develop` with `ConsumerBusyException: Exclusive consumer is already connected` and passes with this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SlavchoGeorgiev:
The second commit (`da4cb87`) is unrelated to this fix — it addresses a pre-existing flake in `TaskSeq tests.Completed tasks come in unpredicted order` that failed one of the two `build` jobs on my first push (the other job passed on the same commit).

The test took two rounds of five picks and asserted the rounds differ. Since `whenAnyTask` picks a *random* completed task, the rounds legitimately match sometimes: I measured 2 collisions in 2000 local trials (~0.1%, i.e. about 1 run in 600). The reported `[1; 1; 1; 1; 1]` on both rounds is the most likely collision pattern, because index 0 holds the never-completing reset task, so a random start index of 0 or 1 both land on the first generator.

The new assertion keeps the same intent — the pick is randomized rather than always the first task — but deterministically: it takes 50 picks and requires more than one distinct task, which only fails if the pick stops being randomized. That moves the flake probability from ~1/600 to ~1e-20. Verified locally: target test 60/60, full unit suite 201/201 over three runs, and the assertion still fails if the threshold is raised, so it isn't vacuous.

Happy to split it into its own PR if you'd prefer this one to stay single-purpose.

tests/IntegrationTests/Basic.fs
```diff
@@ -24,6 +25,46 @@ let tests =
...
+                    .SubscribeAsync()
+
+            // a malformed message crashes the consumer mailbox
+            post (consumer :?> ConsumerImpl<byte[]>).Mb
+                (ConsumerMessage.MessageReceived(struct (Unchecked.defaultof<RawMessage>, Unchecked.defaultof<ClientCnx>)))
+            let rec waitUntilFailed () : Task<unit> =
+                task {
+                    let! isConnected = consumer.IsConnected()
+                    if isConnected then
+                        do! Task.Delay 10
```
Lanayx: In F# recursion on task is not optimized and will keep increasing heap memory, it should either be rewritten to async or I'm fine with increasing to 100ms, but also providing exit after around 5 seconds to avoid test hanging indefinetely

SlavchoGeorgiev: I removed the recursion and added a timeout of around 5 seconds.

